PR #3, opened 2026-02-26
Refactor and handle multi-versioning

aferr:
Here is a proposed refactor to allow our two versions to coexist. The idea is to support two toolchains by having two projects in the same repo. The core includes just the semantics which currently builds with 4.22.0.

The tactics and example programs which genuinely depend on Lean 4.28.0 are moved to a separate package in the same repo under `kraken-experimental`.  The experimental package includes everything and depends on the core semantics.

Since both modules are complete packages with proper toolchain setups the CI is amended to test both work.

Ideally we should be able to update the core semantics to 4.27.0 in a few weeks but it looks like nothing is lost by keeping just that part at 4.22.0 and it does help me.

(Initially I tried this a different way where the directory structure was shuffled around but there were not truly independent toolchain setups. This is much better, I think).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 /.lake
 .vimsession
+# Build artifacts from test runs
+*.o
+asm-tests/test
--- a/Kraken.lean
+++ b/Kraken.lean
@@ -8,6 +8,10 @@ For experimental features (SymM tactics), see kraken-experimental/.
 -/
 
 import Kraken.Semantics
+import Kraken.Parser
+import Kraken.TestHarness
 import Kraken.Theorems
 import Kraken.Tactics
 import Kraken.Examples
+
+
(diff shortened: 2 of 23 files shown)


protz (commented):
I looked at this. Based on our conversation this morning... I think this is where we ended at:
- In Examples.lean, only p11 needs to be in the tacticsv2 directory (i.e. p1, p2, p3 remain in tacticsv1)
- in Tactics.lean, step_one should move to tacticsv1 directory

Then, p1, p2 and p3 can be examples that are tested with an older Lean version (like 4.27), relying on tacticsv1 (which for the time being would only have step_one, and perhaps reg_dec_loop as a reusable tactic?). Eventually you'd have to grow tacticsv1 to something more comprehensive, while we work on tacticsv2.

Does that make sense?

As to the build: is there any way to override the lean-toolchain in GitHub actions? I'd love for the default build to be with v4.28 and have a supplemental action that overrides this to be 4.27 and additionally disables the "tacticsv2" directory. But I don't know if this can be done efficiently.

Let me know your thoughts! Cheers.

protz:
I think the new design is reasonable. Let's not get too hung up on it. We can switch from 4.22 to 4.27 when the time comes for the less experimental part.

PR #20, opened 2026-03-05
Add nop/testb, fix imulq imm and local labels

aferr:
These changes are all motivated by running the kraken semantics against GNU assembler for a 20ish interesting programs.

Parser:
- Standalone labels now emit nop instead of erroring
- Parse errors now include line number and content

Semantics:
- Add nop instruction (no-op)
- Add testb (8-bit test) instruction
- Fix imulq to accept immediate operands

TestHarness:
- Preserve .L* local labels (was incorrectly filtering as directives)

Tests:
- Add testb and imulq immediate test cases

### DIFF
--- a/Kraken/Parser.lean
+++ b/Kraken/Parser.lean
@@ -576,21 +576,6 @@ def skipToEndOfLine : Parser Unit := do
   let _ ← many (satisfy fun c => c != '\n')
   let _ ← (pchar '\n' *> pure ()) <|> pure ()
 
-/-- Parse multiple lines into a Program. -/
-partial def parseProgramAux (acc : Program) : Parser Program := do
-  skipHWs
-  let done ← (eof *> pure true) <|> pure false
-  if done then
-    pure acc
-  else
-    let line ← parseLine
-    skipToEndOfLine
-    match line with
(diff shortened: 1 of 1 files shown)


Kraken/Parser.lean
```diff
@@ -353,23 +355,54 @@ def parseInstr : Parser Instr := do
...
+    let src ← parseOperand; parseComma; let dst ← parseRegOrMem
+    pure (.testb dst src)
 
   -- Shift instructions - 64-bit
   | "shl" | "shlq" | "sal" | "salq" => do
-    let cnt ← parseOperand; parseComma; let dst ← parseRegOrMem
-    pure (.shl dst cnt)
+    -- Support both forms: shlq $count, %dst OR shlq %dst (implicit count=1)
+    let first ← parseOperand
+    let hasComma ← (attempt parseComma *> pure true) <|> pure false
```
andres-erbsen: I am not really on top of parsing in lean, but might there be a way to parse a comma-separated list and then match on it, with cases for the numbers of elements we support?

andres-erbsen (commented):
Definitely good to have more tests, thank you!

PR #25, opened 2026-03-10
Changes for compatibility with a downstream application

aferr:
To simplify building for a downstream application this change modifies some build opitons, notably:
```
autoImplicit = false
relaxedAutoImplicit = false
```
which makes implicit variables (such as universes for types) unavailable. (Another concrete use-case is that the initial state must be explicit in eventually_weaken. That I actually kind of like)

This build change is unlikely to change for the downstream app after the upgrade to 4.27.

How do folks feel about adding these options to the build here? You can see the impact concretely below. The main motivation is to simplify maintaining two separate builds of the codebase, but if keeping implicit type universes is important for people (or others really hate having this) I can workaround not having it here.

### DIFF
--- a/Kraken/Parser.lean
+++ b/Kraken/Parser.lean
@@ -692,4 +692,3 @@ loop:
 #eval parse! "leaq 16(%rbp, %rcx, 4), %rax"
 
 end Tests
-


Kraken/Parser.lean
```diff
@@ -644,31 +644,31 @@ section Tests
 open Kraken.Parser
 
 -- Test: Simple instruction
-#eval parse! "addq %rax, %rbx"
+#eval! parse! "addq %rax, %rbx"
```
eric-wieser: Why did these change? I created #28 since this looked suspicious.

aferr: It's not actually needed and `parse!` already panics on these. So I reverted this aspect. Thanks.

andres-erbsen:
Sure.

Maybe you've already thought of this, but could we use `set_option` inside the concatenated file to toggle this setting while keeping our code as is?

PR #29, opened 2026-03-13
Switch to basic blocks and make control workable for the evaluator

protz:
This is a draft -- but this is what we had in mind when talking about basic blocks.

This now creates more mutual recursion and there's an issue with the mutual partial fixpoints but this is the idea.

I think in this style, evaluation jumps back and forth between basic blocks and control instructions, and returns only when there is no next instruction


protz:
Perhaps the solution to the partial_fixpoint issue is to merge everything into one big function, but that's suboptimal.

Issue #32, opened 2026-03-13
Fix semantics of extra instructions introduced in #9


protz:
I'm working on redoing my register aliasing change, meaning that I am looking at every instruction.

I noticed that the shift instructions do not update the flags correctly. According to https://www.felixcloutier.com/x86/sal:sar:shl:shr

"The OF flag is affected only for 1-bit shifts (see “Description” above); otherwise, it is undefined. The SF, ZF, and PF flags are set according to the result. If the count is 0, the flags are not affected. For a non-zero count, the AF flag is undefined."

PR #33, opened 2026-03-13
A revised version of width-polymorphic semantics, fixes #16

### LINKED ISSUE #16: Type disambiguation of instruction semantics vs different operand codes
Apologies if I misunderstood something. It seems like we're having a mix of two styles for the semantics.

First style, which is currently in the repository. One instruction per width: we have both `add` and `addl`, which correspond to the same `ADD` instruction, but encode different behavior depending on the width of the operands.

Second style, which I advocate in #4:
- the type of the destination operand tells us which variant of the instruction we are looking at
- virtual registers with get/set functions implement the particular behavior of a virtual register (zero-extend or preserve)
- some width-polymorphic functions allow sharing of the code between variants

Example: https://www.felixcloutier.com/x86/mul and the corresponding implementation here: https://github.com/AeneasVerif/kraken/pull/4/changes#diff-46e63f066356de3d5f8271e1e241929704ca4e22e712de2b5270ccfc9664140fR485 where the result of the multiplication can be written in one variant for 16, 32 and 64-bit registers (this PR leaves aside 8-bit registers which would conceivably need a special case).

Another example: https://github.com/AeneasVerif/kraken/pull/4/changes#diff-46e63f066356de3d5f8271e1e241929704ca4e22e712de2b5270ccfc9664140fR425 for ADD which relies on a "natural" sign-extension of the source, along with the polymorphic behavior of reg.set, to keep only a single instruction.

Same deal with `mov`: #4 allows having a single mov instruction, I believe, relying on the type of the destination operand to know which variant we are looking at. (We would then have `movzx` to deal with `mov`s that do zero-extensions).

There's probably a reason why the current style is in the `main` branch but I would like to use this issue to discuss the pros and cons of both styles. My intuition is that it's better to stick closer to the instruction encoding and have fewer variants, and it also allows greater code sharing.

protz:
This takes into the account the discussions in #16 and in #4 and overhauls Semantics.lean.

The main change from the previous version is:
- memory is now annotated with an operand size (and we assume address size is always 64 bits)
- this is enough to determine the variant of instruction that is used in every single case, except for the `push` operation
- for that latter case, I have a `TypedOperand` that adds extra width information to the immediate case

Generally speaking, I tried hard to address code duplication in the current version of the semantics. I noticed that a lot of matches were quite big, and could be simplified with a modest amount of dependent types. I was initially reluctant to introduce dependent types in the semantics, on the basis that it might complicate writing the interpreter, but it turns out that Lean was quite good at things and I was never in dependent type hell where types don't reduce and I have to sprinkle annotations and/or manually do type conversions. Look at `Registers.get` and `Registers.set` and notice how much shorter this is now.

A nice side-effect of sprinkling dependent types is that a few operations that were previously throwing are now total. We could probably make a lot more operations direct (instead of CPS) by filling out readMem, setMem and so on for non-aligned cases, which in turn could make effective_addr non-CPS, provided the caller proves that the operand is a `.mem`. This would require changing the type of `.lea` to take exactly a register and exactly a memory. This is fine, and I've done it for a few other cases. We may want to address this soon (but as a separate issue) as it may simplify things and/or influence how proofs are written against those semantics.

The main difficulty of this PR is going to be proofreading the semantics. I noticed several errors in the shifts and rotates (unless I misunderstood the manual), so these should be assumed to be incorrect and we'll track those via #32 -- let's leave it aside for now. But I'd be particularly interested in feedback on {add,sub}_{overflow_,}_with_carry, as well as the rest of the semantics, notably .mov and .movzx

@aferr I did not fix the parser -- for now, I'd like to get feedback on the Semantics. Once we're good, I'm hoping you can help me fix the parser since you're familiar with that part of the code. Thanks.

### DIFF
--- a/Kraken/Semantics.lean
+++ b/Kraken/Semantics.lean
@@ -11,12 +11,6 @@ For tactics, see Kraken/Tactics.lean.
 import Std
 import Lean.Elab.Tactic.Grind
 
-class Throw α where
-  throw: String → α
-
-def throw {α} [inst: Throw α] :=
-  inst.throw
-
 -- ============================================================================
 -- Width Type
 -- ============================================================================
@@ -493,6 +487,12 @@ def MachineState.getReg (s : MachineState) (r : Reg) : UInt64 :=
 
 def MachineState.setReg (s : MachineState) (r : Reg) (v : UInt64) : MachineState :=
   { s with regs := s.regs.set r v }
+
+class Throw α where
+  throw: String → α
+
+def throw {α} [inst: Throw α] :=
+  inst.throw
 
 def MachineState.readMem [Throw α] (s : MachineState) (addr : Address) (width: Width) (ret: Word → α): α :=
   if addr % 8 != 0 then


Kraken/Semantics.lean
```diff
@@ -277,87 +419,47 @@ def Instr.is_ctrl
...
-@[simp] def Registers.getRaw (regs : Registers) (r : Reg) : UInt64 :=
-  match r.base with
+def Registers.get64 (regs: Registers) (r: BaseReg): UInt64 :=
+  let ⟨ r, _ ⟩ := r
+  match r with
   | .rax => regs.rax | .rbx => regs.rbx | .rcx => regs.rcx | .rdx => regs.rdx
   | .rsi => regs.rsi | .rdi => regs.rdi | .rsp => regs.rsp | .rbp => regs.rbp
   | .r8  => regs.r8  | .r9  => regs.r9  | .r10 => regs.r10 | .r11 => regs.r11
   | .r12 => regs.r12 | .r13 => regs.r13 | .r14 => regs.r14 | .r15 => regs.r15
-  | _ => 0  -- Unreachable for base registers
```
protz: this and the default case below (with the "unreachable" comment) are now gone thanks to the use of a refinement type -- Lean is quite good at doing the match exhaustiveness analysis automatically, I was very pleasantly surprised

Kraken/Semantics.lean
```diff
@@ -475,75 +540,81 @@ def eval_operand [Throw α] (s : MachineState) (o : Operand) (ret: UInt64 → α
...
+  | .mem w _ _ _ _ => effective_addr s o (fun addr => s.readMem addr w ret)
 
-def eval_reg_or_mem [Throw α] (s : MachineState) (o : Operand) (ret: UInt64 → α): α :=
+def eval_typed_operand [Throw α] (s: MachineState) (o: TypedOperand) (ret: UInt64 → Width → α): α :=
   match o with
-  | .reg r => ret (s.getReg r)
-  | .mem _ _ _ _ => effective_addr s o (fun addr => s.readMem addr ret)
+  | .typed (.imm v) w => ret (eval_imm v) w
+  | .typed (.reg r) () => ret (s.getReg r) r.width
+  | .typed (.mem w base idx scale disp) () => effective_addr s (.mem w base idx scale disp) (fun addr => s.readMem addr w (fun v => ret v w))
```
protz: are there `as`-patterns in lean4? I could not figure out the syntax and ended up rebuilding the `.mem` case as the argument to `effective_addr` -- @eric-wieser

Kraken/Semantics.lean
```diff
@@ -73,16 +91,120 @@ inductive Reg
...
-@[simp] def Reg.base : Reg → Reg
-  | rax | eax | ax | al => .rax | rbx | ebx | bx | bl => .rbx
-  | rcx | ecx | cx | cl => .rcx | rdx | edx | dx | dl => .rdx
-  | rsi | esi | si | sil => .rsi | rdi | edi | di | dil => .rdi
-  | rsp | esp | sp | spl => .rsp | rbp | ebp | bp | bpl => .rbp
-  | r8  | r8d  | r8w  | r8b  => .r8  | r9  | r9d  | r9w  | r9b  => .r9
-  | r10 | r10d | r10w | r10b => .r10 | r11 | r11d | r11w | r11b => .r11
-  | r12 | r12d | r12w | r12b => .r12 | r13 | r13d | r13w | r13b => .r13
-  | r14 | r14d | r14w | r14b => .r14 | r15 | r15d | r15w | r15b => .r15
+@[simp] def Reg.is_base : Reg → Prop
```
eric-wieser: Note that the Lean style is to name propositions as
```suggestion
@[simp] def Reg.IsBase : Reg → Prop
```

Kraken/Semantics.lean
```diff
@@ -135,17 +257,49 @@ deriving Repr
...
   -- Standard x86: base + idx*scale + disp. E.g. 8(%rsp) = disp 8, (%rsi,%r15,8) = idx .r15
   -- Per Intel SDM Vol. 2A Section 2.1.5 (SIB byte), valid scale values are 1, 2, 4, 8.
   -- The default scale is 1 (SIB SS bits = 00). Scale must be explicit in AT&T syntax when != 1.
 deriving Repr, BEq
 
+-- There is one case where the variant of instruction cannot be determined --
+-- pushing an immediate to the stack. For that case, we annotate the immediate
+-- with a width.
+inductive TypedOperand
+| typed (o: Operand) (extra: if o matches Operand.imm _ then Width else Unit)
```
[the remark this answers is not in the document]
eric-wieser:
```suggestion
| typed (o: Operand) (extra: match o with | Operand.imm _ => Width | _ => Unit
```
saves you an intermediate bool conversion

PR #39, opened 2026-03-24
factored semantics

andres-erbsen:
This PR fills in a number of TODOs in the semantics, including width-parametrized operations, high-byte registers, and computed control flow. It needs work -- only the Semantics file builds.

### DIFF
--- a/Kraken/Parser.lean
+++ b/Kraken/Parser.lean
@@ -210,7 +210,7 @@ def parseMemory : Parser AddrExpr := do
   let _ ← pchar ')'
   -- Some adapters between the parsed components and the expected dependent
   -- pairs:
-  let idx := Option.map (fun idx => (idx, scale)) idx
+  let idx := Option.map (fun idx => ⟨idx, scale⟩) idx
   pure ({ base, idx, disp })
 
 def parseImm w : Parser (Operand w) := do
@@ -902,7 +902,7 @@ info: [Directive.Instr
                     (Operand.RegOrMem
                       (RegOrMem.mem
                         { base := some ⟨Width.W64, RegOrRip.Reg (Reg.low (Reg64.rsi) (Width.W64))⟩,
-                          idx := some (⟨Width.W64, Reg.low (Reg64.r15) (Width.W64)⟩, Width.W64),
+                          idx := some { reg := ⟨Width.W64, Reg.low (Reg64.r15) (Width.W64)⟩, scale := Width.W64 },
                           disp := ConstExpr.Int64 0 })) }]
 -/
 #guard_msgs in
@@ -1010,7 +1010,7 @@ info: [Directive.Instr
      operation := Operation.lea
                     (Reg.low (Reg64.rax) (Width.W64))
                     { base := some ⟨Width.W64, RegOrRip.Reg (Reg.low (Reg64.rbp) (Width.W64))⟩,
-                      idx := some (⟨Width.W64, Reg.low (Reg64.rcx) (Width.W64)⟩, Width.W32),
+                      idx := some { reg := ⟨Width.W64, Reg.low (Reg64.rcx) (Width.W64)⟩, scale := Width.W32 },
                       disp := ConstExpr.Int64 16 } }]
 -/
 #guard_msgs in
--- a/Kraken/Semantics.lean
+++ b/Kraken/Semantics.lean
@@ -225,9 +225,14 @@ instance: DecidableEq ((w: Width) × Reg w) := by
   . apply isFalse
     simp_all
 
+structure AddrIndex where
+  reg: Σ w, Reg w
+  scale: Width
+deriving Repr, BEq, DecidableEq
+
 structure AddrExpr where
   base : Option (Σ w, RegOrRip w)
-  idx : Option ((Σ w, Reg w) × Width)
+  idx : Option AddrIndex
   disp : ConstExpr := .Int64 0
 deriving Repr, BEq, DecidableEq
 
@@ -242,7 +247,7 @@ def AddrExpr.interp [Layout] [address_size : AddressSize] (a : AddrExpr) (s : Re
               | .some ⟨_, .Rip⟩ => (layout p.next).toInt
               | .none => 0
   let idx := match a.idx with
-             | .some (⟨_, r⟩, c) => (s.get r).toInt * c.bytes
+             | .some ⟨⟨_, r⟩, c⟩ => (s.get r).toInt * c.bytes
              | .none => 0
   BitVec.ofInt address_size.address_size.bits (base + idx + (a.disp.interp p).toInt)
 


Kraken/Semantics.lean
```diff
@@ -104,262 +61,15 @@ structure Registers where
...
-
-/-- Set a register value with appropriate aliasing behavior:
-    - 64-bit: direct write
-    - 32-bit: zero-extends to 64-bit (clears upper 32 bits) per Intel SDM
-    - 16-bit: preserves upper 48 bits
-    - 8-bit: preserves upper 56 bits -/
-def Registers.set (regs : Registers) (r : Reg) (v : UInt64) : Registers :=
-  match r with
-  -- 64-bit registers: direct write
+def Reg64s.set64 (regs : Reg64s) (r : Reg64) (v : Width.W64.type) (v := UInt64.ofBitVec v) : Reg64s := match r with
```
protz: as discussed: this is an optional argument, so unless you envision callers occasionally passing a bitvec that is *not* using `v`, let's just use a regular let-binding

Kraken/Semantics.lean
```diff
@@ -368,732 +78,750 @@ def Registers.set (regs : Registers) (r : Reg) (v : UInt64) : Registers :=
...
+  else match s.dmem[UInt64.ofBitVec addr]? with
+  | .some v => ret (v.toBitVec.truncate _)
+  | .none => throw (s!"Memory accessed but not mapped (addr={repr addr})")
 
--- Sign-extension helpers: use standard integer type conversions
--- Strategy: truncate to input size → signed Int conversion → convert to UInt64
--- This avoids branching on sign bit and is more proof-friendly.
+def MachineData.store [Throw α] (s : MachineData) (addr : BitVec 64) {w : Width} (v : w.type) (ret: MachineData → α) : α :=
+  s.load addr .W64 (fun old =>
+  ret { s with dmem := s.dmem.insert (.ofBitVec addr) (.ofBitVec (old.replaceLow v)) })
```
protz: neat

Kraken/Semantics.lean
```diff
@@ -368,732 +78,750 @@ def Registers.set (regs : Registers) (r : Reg) (v : UInt64) : Registers :=
...
+    let s := s.setReg r_hi (.ofNat _ (v >>> w.bits))
+    next s)
+  | .imul dst src1 src2 =>
+    src1.interp s p (fun a =>
+    src2.interp s p (fun b =>
+    let v := a * b
+    s.set (match (generalizing := false) (motive := Option (RegOrMem w) → RegOrMem w)
+             dst with | .some dst => dst | _ => src1) v p (fun s =>
+    let cf := v.toInt != a.toInt * b.toInt
+    undefined (λ sf => undefined (λ zf => undefined (λ af => undefined (λ pf =>
```
protz: do you want to define `undefined4` or something?

Kraken/Parser.lean
```diff
@@ -169,57 +189,73 @@ def parseMemory : Parser Operand := do
...
+    | '$' => parseInt64
+    | '.' => parseLabel
+    | _ => fail "not an immediate"
+  pure (.imm i)
+
+-- We need to eagerly parse (to move through the syntax), but we may need to
+-- defer choosing a width for those operands that are untyped (as in: may have
+-- any width).
+def MaybeTyped (T: Width → Type) :=
+  Σ (w: Option Width), match w with | .some w => T w | .none => (w: Width) → T w
```
eric-wieser: I'm curious, does lean let you drop the `fun _ =>`s below if you write
```suggestion
  Σ (w: Option Width), match w with | .some w => T w | .none => {w: Width} → T w
```
(this is purely a stylistic change, whether it's worth it depends on whether it makes later code cleaner or less clean)

Kraken/Parser.lean
```diff
@@ -254,128 +353,220 @@ def parseInstr : Parser Instr := do
...
 
   -- Shift instructions - 64-bit
   | "shl" | "shlq" | "sal" | "salq" => do
     let cnt ← parseOperand; parseComma; let dst ← parseRegOrMem
     pure (.shl dst cnt)
+
   | "shll" | "sall" => do
     let cnt ← parseOperand; parseComma; let dst ← parseRegOrMem
     pure (.shll dst cnt)
+
```
eric-wieser: Note that the typical lean style is to avoid blank lines within definitions, though that's probably in the space of things you could decide to ignore.

Kraken/Semantics.lean
```diff
@@ -368,732 +82,798 @@ def Registers.set (regs : Registers) (r : Reg) (v : UInt64) : Registers :=
...
-  else
-    ret { s with heap := s.heap.insert addr val }
+def Reg.interp {w} (r : Reg w) (s : MachineData) (_ : Position) (ret : w.type → α) :=
+  ret (s.regs.get r)
 
--- Sign-extension helpers: use standard integer type conversions
--- Strategy: truncate to input size → signed Int conversion → convert to UInt64
--- This avoids branching on sign bit and is more proof-friendly.
+def MachineData.load [Throw α] (s : MachineData) (addr : BitVec 64) (w : Width) (ret : w.type → α): α :=
+  if addr % 8 != 0 then throw (s!"Unimplemented: only 8-byte-aligned memory access is supported")
```
eric-wieser: nit:
```suggestion
  if addr % 8 != 0 then throw s!"Unimplemented: only 8-byte-aligned memory access is supported"
```

Kraken/Semantics.lean
```diff
@@ -368,732 +82,798 @@ def Registers.set (regs : Registers) (r : Reg) (v : UInt64) : Registers :=
...
+  rcases p2 with ⟨ w2, r2 ⟩
+  by_cases h: w1 = w2
+  . subst h -- rw leaves a heterogeneous equality
+    by_cases h2: r1 = r2
+    . apply isTrue
+      simp_all
+    . apply isFalse
+      simp_all
+  . apply isFalse
+    simp_all
```
eric-wieser: I'm pretty sure `inferInstance` will find all of these, and so they can all just be deleted?

Kraken/Semantics.lean
```diff
@@ -1,93 +1,54 @@
...
+def BitVec.drop (x : BitVec w) (n : Nat) : BitVec (w - n) := x.extractLsb' n (w-n)
+def BitVec.replaceLow (old : BitVec w) (new : BitVec n) : BitVec w :=
+  (BitVec.append (old.drop w) new).setWidth _
 
--- ============================================================================
--- Width Type
--- ============================================================================
+inductive Width | W8 | W16 | W32 | W64 deriving Repr, BEq, DecidableEq
 
-/-- Operand width for multi-width instructions. -/
```
eric-wieser: Let's keep the docstrings where they still apply!

protz:
Question for @eric-wieser: is there a leanfmt you recommend?

protz (approved):
We will merge this and I will file a number of followup issues -- this has been in flight for a long while.

PR #62, opened 2026-04-08
Parser Changes to Support Some Programs

aferr:
This PR batches changes to support some specific programs. These changes affect only the parser.

The changes are as follows:
1) Supporting AT&T Suffix variants for mul which are mulq, mull, mulw, mulb
2) Support a syntax GNU AS supports for single operand shifts, which is a shorthand for shifting by 1
3) Fix whitespace -- it is permitted for a space to come between a displacement and an open parenthesis

A comment is added to explain that the GNU documentation is the source of truth.

More details on these changes:

----------------------------
Suffixes:
----------------------------
Their implementation here is based off of:
- GNU AS, which is the de facto ref manual for AT&T syntax AFAICT
and especially the suffix descriptions
https://sourceware.org/binutils/docs/as/i386_002dMnemonics.html#Instruction-Naming
- the implementation of other suffixed instructions in the Parser

Also adds GNU AS docs as a link in a comment at the top declaring it
is the primary reference for this file.

----------------------------
Single Operand Shifts:
----------------------------
GNU AS permits writing single argument shift as a shorthand
for shift by 1. Here is an example:
```
  shl %r9
```

I cannot find a document that explains this encoding but running GNU on this
instruction validates that this is permitted by GNU and it has this
functionality.

Here is a test you can use to confirm this:
```
.global main
.text main:
  mov $42, %r9
  shr %r9
  mov %r9, %rax
  ret
```
which will return exit code 21 (shifting 42 right by 1).

It appears supported for shifts and rotates generally, so this also
adds support for this syntax for other shifts and rotates using the same
solution.

----------------------------
Whitespace fix
----------------------------
Tolerate a space between displacements and open parenthesis as in:
```
mov %rax, 0 (%rdi)
```

where the space after the 0 is the important part that this change allows

### DIFF
--- a/Kraken/Parser.lean
+++ b/Kraken/Parser.lean
@@ -290,6 +290,7 @@ def parseShiftExpr: Parser ShiftCountExpr := do
     let _ ← pstring "%cl"
     pure .cl)
 
+
 -- ============================================================================
 -- Condition Code Parsing
 -- ============================================================================
@@ -314,6 +315,10 @@ def parseComma : Parser Unit := do
   skipHWs
   let _ ← pchar ','
   skipHWs
+
+/-- Try to parse a shift count followed by a comma; if that fails, default to 1. -/
+def parseOptionalShiftAndComma : Parser ShiftCountExpr :=
+  (attempt do let cnt ← parseShiftExpr; parseComma; pure cnt) <|> pure (.imm8 (.Int64 1))
 
 def ascribe {T: Width → Type} (w: Width) (v: MaybeTyped T): Parser (T w) := do
   match v with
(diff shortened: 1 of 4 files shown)


protz:
My bad -- I only looked at the CI output.

Let's try to fix these in a separate PR that also adds these tests on CI to prevent future regressions.

protz (changes requested):
Looks good!

- Please hoist the definition for parseOptionalShiftAndComma, we do not want 10 copies of the same definition.
- I agree with Miriam that we want a test for this, even if other tests are currently broken.

Can we add the test now to the gas test suite with the understanding that it'll be enabled on CI once we fix the other regressions?

Thank you!

aferr:
Refactor done and new tests added. You should be able to run `scripts/run_all_asm_tests.sh` and see that the 3 new tests pass. I also see that if I do `git checkout main -- Kraken/Parser.lean` and run these 3 tests after that, they will fail.

miriampolzer:
Great, thank you for checking and adding the tests! I'd recommend we fix the existing tests as soon as possible so we don't end up with more regressions (in a follow-up or parallel PR -- I think Jonathan is working on this).

Even with a CI in place, I find it really helpful to run relevant tests manually before committing as it gives much faster feedback than waiting for a CI build. I think it would be a useful habit for us to adopt for Kraken going forward to keep the main branch clean.

protz (approved):

protz:
Thank you! I'll merge this into my branch

PR #63, opened 2026-04-08
Parser line nums for errs; refactor parser tests

aferr:
This change does two things:
1) Report line numbers from the Parser when there are errors
2) Move tests out of the Parser and into a separate file.


protz (approved):
This looks generally ok to me.

The current version of `parse` is because Andres and I were attempting to turn `parse` into something that can be evaluated with reduction for our test purposes, and `parseProgram` was calling a `partial` def that was blocking reduction (from what I remember).

The original version is this:

```
-/-- Parse an assembly string into a Program.
-    Returns an error message on failure. -/
-def parse (input : String) : Except String Program :=
-  match parseProgram ⟨ input, input.startPos ⟩ with
-  | .success _ prog => .ok prog
-  | .error _ .eof => .error "unexpected end-of-file"
```

but since we want to keep track of line numbers, I don't think there's any benefit to reverting to the earlier, more idiomatic version.

miriampolzer:
I would agree with removing these in-line tests, but make sure their content is covered by the existing assembly tests or create new ones before.

aferr:
> but since we want to keep track of line numbers, I don't think there's any benefit to reverting to the earlier, more idiomatic version.


I have no problem with your way of defining it. I just wanted line numbers in the error messages (which we also did not have before :) )

aferr:
> I would agree with removing these in-line tests, but make sure their content is covered by the existing assembly tests or create new ones before.

That's a fair request! I think I don't have time to get to that right now. Would you still support splitting these off to a separate file as a temporary measure before we port these to assembly tests?

miriampolzer:
Yes sure! Splitting up tests and program logic is a good step regardless.

protz:
> I would agree with removing these in-line tests, but make sure their content is covered by the existing assembly tests or create new ones before.

the reason these inline tests were there is for debugging the parser -- I found it easier to have little tidbits directly processed through `#eval` rather than have a test harness with a main (since that's potentially more stuff to debug)

now that the parser has stabilized (especially with #65) I agree that these are not as necessary as they used to

aferr:
Ah I thought this was perhaps a bit crufty when I added those inline in the same file initially https://github.com/AeneasVerif/kraken/commit/7e2c98d8f233b5696cec3b66c19f5c2706706f5d#diff-5e5bdcb7fd5ecfee128968274baabfac5669a2c8ab953c18d341c95ce1fe5057R396

But I'm glad they remained helpful for a while

Issue #30, opened 2026-03-13
GNU AS tests in CI?


aferr:
Should we run (some of?) the tests in `asm-tests` with `Kraken/TestHarness.lean` in CI?

protz:
This is #65 I believe

aferr:
I agree

PR #65, opened 2026-04-09
Start cleaning up tests: share prologue / epilogue

protz:
Prompted by https://github.com/AeneasVerif/kraken/pull/62 I started cleaning up the tests. 
- Add support for working on OSX via emulation
- Share prologue / epilogue in tests rather than copy-paste
- Make krakentest support both _start and _main in its test harness

### DIFF
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -30,6 +30,5 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: leanprover/lean-action@v1
-        with:
-          test-cmd: lake build krakentest
+      - run: lake build krakentest
       - run: ./scripts/run_all_asm_tests.sh
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -12,6 +12,8 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: leanprover/lean-action@v1
+      - name: Build Kraken Test
+        run: lake build krakentest
 
   build-experimental:
     name: Build KrakenExp (4.28.0)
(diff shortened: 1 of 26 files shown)


protz:
Interestingly enough, it seems like for OSX, program execution does not start with registers being zeroed out, so my OSX prologue needs to reset all of these registers.

protz:
With this patch, I am at:

```
Running test_32bit_ops... ✗
Running test_8bit_16bit... ✗
Running test_arithmetic... ✗ (execution error)
Running test_bswap... ✗
Running test_conditionals... ✗
Running test_flags... ✓
Running test_logic... ✓
Running test_memory_whitespace... ✗
Running test_memory... ✗
Running test_move... ✗
Running test_multiply... ✗
Running test_shifts... ✓
Running test_zero_extend... ✗
```

so perhaps I'll keep debugging why my tests fail. This is fewer passing tests than what @aferr reports on #62.

protz:
A lot of my tests seem to disagree on the expected final value of the flags:

```
FAIL:

  PF: expected false, got true
  ZF: expected false, got true
```

so perhaps i need to figure out how to reset the flags in the prologue, just like I do with registers.

protz:
Ok with this I now ensure that we know what the initial state of all flags and registers is, on both linux and osx.

This brings me to the same state as #62 so that I can start helping out with test debugging.

protz:
The failure on CI is legitimate.

> test_32bit_ops: FAIL:
> AF: expected false, got true

test relies on unspecified value of AF flag after a shift (`shl` instruction)

> test_mul_suffixed: FAIL:
> PF: expected false, got true
> ZF: expected false, got true

test relies on unspecified values for these flags after a `mul` ("The SF, ZF, AF, and PF flags are undefined.")

> test_multiply: FAIL:
> PF: expected false, got true
> ZF: expected false, got true

same deal with `imul`

> test_shifts: FAIL:
> AF: expected false, got true

`shr` (AF undefined) followed by other instructions that preserve AF

aferr:
Closes #30

miriampolzer:
Just talked Jonathan offline, I will pick this up and spin off a couple of PRs:
1. refactor the test logic for decuplication, correctness, and maintainability (draft from yesterday: https://github.com/AeneasVerif/kraken/commit/80482438751921a92478e22f33813d109edc2eb1)
2. test fixes from this PR (parser, semantics)
3. allow tests to skip checking some flags via metadata (to fix remaining tests)
4. add tests to CI (linux only)
5. add mac support and add mac to CI (including checks and additional flags added here to skip tests if operations are not supported)

As followup I would like to split the tests up to make them proper unit tests. Right now they do a lot of things, and when a test fails it is not straightforward to see which instruction caused the issue.

protz:
Most of this was merged as part of #70. What remains in here is the OSX support logic, for which I will file an issue. Thanks @miriampolzer!

PR #66, opened 2026-04-10
Permit Projects to Depend on Both Kraken and Mathlib 4.27 Simultaneously

aferr:
This is to resolve incompatibility issues between these definitions and Mathlib 4.27 which also defines its own scanl. With this change one can depend on both Kraken and Mathlib 4.27, and this was not possible without this change.

### DIFF
--- a/Kraken/Semantics.lean
+++ b/Kraken/Semantics.lean
@@ -638,8 +638,10 @@ def Layout.apply (l : Layout) (prog : Program) : Executable :=
   (l.start, prog.mapIdx (fun i d => (d, l.size i)))
 instance : CoeFun Layout (fun _ => Program → Executable) where coe := Layout.apply
 
--- TEMPORARY: replace with `import Init.Data.List.scan.Basic` when dropping support for Lean 4.28
-namespace List
+-- TEMPORARY: delete when dropping support for Lean < 4.28 (which adds Init.Data.List.scan.Basic)
+-- This permits use of Mathlib 4.27 which also implements its own `scanl` which differs
+-- from the one in `List.Basic` in Lean 4.28
+namespace Kraken.Compat
 @[inline]
 private def scanAuxM {α β m} [Monad m] (f : β → α → m β) (init : β) (l : List α) : m (List β) :=
   go l init []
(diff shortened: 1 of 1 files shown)


aferr:
To see that this is really necessary one can attempt to build this branch which is based off the same commit as this one: https://github.com/AeneasVerif/kraken/commit/b509cec2de11b46550dd8c069fc4c1a45965f7b3

On that branch one can run:
```
lake exec cache get
lake build
lake env lean Test/MathlibCompat.lean
```

Which should result in something like:
```
Test/MathlibCompat.lean:3:0: error: import Batteries.Data.List.Basic failed, environment already contains 'List.scanl' from Kraken.Semantics
```

The need here is demonstrated by MathlibCompat.lean -- it's useful to depend on both Kraken and Mathlib 4.27

I am open-minded to other ways of meeting this need though!

Kraken/Semantics.lean
```diff
@@ -638,8 +638,10 @@ def Layout.apply (l : Layout) (prog : Program) : Executable :=
   (l.start, prog.mapIdx (fun i d => (d, l.size i)))
 instance : CoeFun Layout (fun _ => Program → Executable) where coe := Layout.apply
 
--- TEMPORARY: replace with `import Init.Data.List.scan.Basic` when dropping support for Lean 4.28
-namespace List
+-- TEMPORARY: delete when dropping support for Lean < 4.28 (which adds Init.Data.List.scan.Basic)
```
aferr: I was confused by the old comment since I think 4.28 is essentially dropped currently. I tried deleting these definitions in the context where both Kraken and Mathlib 4.27 are imported and that also did not work. So I thought the comment was supposed to mean what is now written here. Did I get that right?

andres-erbsen: I meant <=4.28.

protz: > since I think 4.28 is essentially dropped currently

I thought we had a CI job that ran both 4.27 and 4.28?

> I meant <=4.28.

@aferr do you mind fixing the comment before merging? thanks!

protz:
Fine by me, as long as we fix the comment (see my remark above). Thanks!

Kraken/Semantics.lean
aferr: Of course will do

aferr: Ah I did not address this point:

> I thought we had a CI job that ran both 4.27 and 4.28?

You are right. What I meant was -- we used to factor the project into 4.22 and 4.28 pieces and this factoring is now gone (and the formerly 4.22 part has been updated). My mental model was then that this is a 4.27 project globally, but I suppose lean may have breaking updates so really we build for both presently.

protz (approved):

PR #68, opened 2026-04-10
Refactor effectful parse into separate file.

aferr:
This is purely for compatibility with a downstream application that cannot handle lean files that contain effectful definitions with `!`.


aferr:
It's possible for me to workaround not having this refactor. Happy to expand on what this does for me in another channel! Let me know your thoughts on this.

aferr:
I will investigate this, discuss elsewhere, and reopen this later since this will take me some time.

aferr:
I think I don't need this now actually. Will just reopen if I need to.

PR #70, opened 2026-04-14
Refactor test running logic

miriampolzer:
Major overhaul of the test runner that checks kraken against hardware. Summary of changes:
- Instead of TestHarness use a small KrakenRunner lean file for running Kraken semantics. It outputs json (regs and flags).
- Move most testing logic to a python file that picks up the json from KrakenRunner and constructs and runs the assembly on hardware.
- Testing boilerplate code is now inside python for easier construction and tracking.


samuelgruetter:
Not sure if this is the right place to post it, but while working on a draft for https://github.com/AeneasVerif/kraken/issues/60, I stumbled on a suspicious line in the semantics of ADC:

https://github.com/AeneasVerif/kraken/blob/694ed68a46930063b9fd24657595d08b4d1a2788/Kraken/Semantics.lean#L368-L372

The last line adds the carry flag twice instead of once. Did you already catch this? (I guess this PR is only about the harness, and not about how thoroughly each instruction is tested, so probably no?)

Kraken/Test/asm_tests.py
```diff
@@ -0,0 +1,171 @@
...
+
+class Color:
+    GREEN = "\033[92m"
+    RED = "\033[91m"
+    BOLD = "\033[1m"
+    RESET = "\033[0m"
+
+def get_boilerplate(instruction_text: str) -> str:
+  reg_count = len(REGS)
+  total_bytes = (reg_count + 1) * 8
```
protz, line 31:
```suggestion
  # We move all base registers + the eflags register into memory, so as to dump it later to stdout
  total_bytes = (reg_count + 1) * 8
```

Kraken/Test/asm_tests.py
```diff
@@ -0,0 +1,171 @@
...
+_start:
+# --- Test Code Start ---
+{instruction_text}
+# --- Test Code End ---
+{moves}
+    pushfq
+    popq %rax
+    movq %rax, _final_state + {reg_count * 8}(%rip)
+
+    movq $1, %rax
```
protz, line 53:
```suggestion
    # print syscall: arguments are 1 (syscall number), 1 (stdout), address of _final_state, and length of _final_state
    movq $1, %rax
```

miriampolzer:
> Not sure if this is the right place to post it, but while working on a draft for #60, I stumbled on a suspicious line in the semantics of ADC:
> 
> https://github.com/AeneasVerif/kraken/blob/694ed68a46930063b9fd24657595d08b4d1a2788/Kraken/Semantics.lean#L368-L372
> 
> The last line adds the carry flag twice instead of once. Did you already catch this? (I guess this PR is only about the harness, and not about how thoroughly each instruction is tested, so probably no?)

Thanks! I don't think our tests had caught this yet, no. Will write a test for it an fix.

Edit: https://github.com/AeneasVerif/kraken/pull/71 to be merged after this one (I marked it as draft).

protz (approved):

KrakenRunner.lean
```diff
@@ -0,0 +1,72 @@
...
+
+def summarize (s : MachineData) : StateSummary :=
+  let r := s.regs
+  let f := s.status
+  { regs := [("rax", r.rax), ("rbx", r.rbx), ("rcx", r.rcx), ("rdx", r.rdx),
+             ("rsi", r.rsi), ("rdi", r.rdi), ("rbp", r.rbp), ("r8", r.r8),
+             ("r9", r.r9), ("r10", r.r10), ("r11", r.r11), ("r12", r.r12),
+             ("r13", r.r13), ("r14", r.r14), ("r15", r.r15)],
+    flags := [("cf", f.cf), ("pf", f.pf), ("af", f.af),
+              ("zf", f.zf), ("sf", f.sf), ("of", f.of)] }
```
eric-wieser, line 45: Is `rsp` deliberately skipped?

samuelgruetter: My guess would be that if you add rsp, all tests would fail, because the compiled runs get a stack pointer determined by address space layout randomization (ASLR), whereas the Lean Semantics runs get a hardcoded one. Did you try that out?

miriampolzer: Yes, I skipped it because it wont match up with the real runs anyways, but the py also has a check to not compare rsp, so adding it here is safe and would not break tests.
Should definitely either add a comment about why it is skipped, or add it back as eric did in his PR, to avoid anyone tripping on this in the future :)

PR #73, opened 2026-04-15
Add not, fix partial writes, parser corrections

miriampolzer:
This is an unchanged copy of the part of https://github.com/AeneasVerif/kraken/pull/65/ that fixed tests. Locally, they are all passing for me now, but as mentioned in the PR there will probably still be some failures in the CI / on other architectures regarding flags that are undefined after the test.

@protz are you happy with these as they are, or would like me to do some changes before merging?

### DIFF
--- a/Kraken/Semantics.lean
+++ b/Kraken/Semantics.lean
@@ -12,7 +12,7 @@ instance : Coe Bool Nat where coe := Bool.toNat
 def BitVec.take {w} (x : BitVec w) (n : Nat) : BitVec n := x.extractLsb' 0 n
 def BitVec.drop {w} (x : BitVec w) (n : Nat) : BitVec (w - n) := x.extractLsb' n (w-n)
 def BitVec.replaceLow {w n} (old : BitVec w) (new : BitVec n) : BitVec w :=
-  (BitVec.append (old.drop w) new).setWidth _
+  (BitVec.append (old.drop n) new).setWidth _
 
 inductive Width | W8 | W16 | W32 | W64 deriving Repr, BEq, DecidableEq, Hashable, Lean.ToExpr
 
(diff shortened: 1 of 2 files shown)


protz (approved):

Kraken/Semantics.lean
```diff
@@ -123,15 +123,32 @@ def throw {α} [inst: Throw α] :=
...
 def MachineData.store {α} [Throw α] (s : MachineData) (addr : BitVec 64) {w : Width} (v : w.type) (ret: MachineData → α) : α :=
-  s.load addr .W64 (fun old =>
-  ret { s with dmem := s.dmem.insert (.ofBitVec addr) (.ofBitVec (old.replaceLow v)) })
+    s.loadOpt addr (fun old =>
+    match old with
+    | .some old =>
+      ret { s with dmem := s.dmem.insert (.ofBitVec addr) (.ofBitVec (old.toBitVec.replaceLow v)) }
+    | .none =>
+      if h: w = .W64 then
+        -- We know how to perform full writes even though there is not previous
```
samuelgruetter, line 145: This differs from bedrock2 semantics. In bedrock2, None means that this address is unmapped and writing to it is forbidden. @andres-erbsen do you want to change this?

samuelgruetter: (and btw, disagreement over this could be an indicator that we should abstract over it, so that each user can pick the store semantics they prefer, so I shamelessly use this an opportunity to campaign for #60 / #72 :wink:)

protz: then we'd need three cases, to express known value / unknown value / forbidden address?

samuelgruetter: In bedrock2, we don't have a third case for "unknown value", and instead use universal quantification.

miriampolzer: I have reverted this part of the change and made some adaptions to the test instead, see https://github.com/AeneasVerif/kraken/pull/76

PR #75, opened 2026-04-15
p4 example trying to preserve sharing using simp

andres-erbsen:


protz (approved):
Discussed with @andres-erbsen -- this is generally useful as something that we aim for with SymM

PR #76, opened 2026-04-16
Add stack to testing, revert write to undef memory

miriampolzer:
Revert the change that allowed writes to memory not already mapped in dmem. As far as I understood the conversation, we consider only memory mapped in dmem accesible to the Program.

For proofs this means if we want to argue about memory that may be uninitialized, we put it in dmem and quantify universally over the content.

For tests we cannot do quantification, but we can just give the program a stack frame (and later also heap space - tbd). For now I added 100B of stack, and separated out the stack part of the test that was failing, as the test was running various unrelated things.

Also add support for subfolders in tests. We have strictly one test per file and we want to write small tests, so we will need subfolders to keep tests in order.


miriampolzer:
Partial revert and test fixes as per discussion in our team meeting yesterday and in https://github.com/AeneasVerif/kraken/pull/73

Let me know if this makes sense, I am still learning how the Semantics are intended, and happy to make changes.

protz (approved):

PR #79, opened 2026-04-16
refactor: rename inductive constructors to follow the naming convention

eric-wieser:
Only types should be in `PascalCase`, everything else (that is not a proof) should be `lowerCamel`.

I've left the width `.W64` ones alone, since there are often exceptions allowed for single-letter identifiers.


protz (approved):
Thanks! I'm happy with aligning stuff to be closer to Lean conventions

PR #80, opened 2026-04-16
refactor: replace `eval% parse! "asm"` with `parse("asm")`

eric-wieser:
The latter better surfaces errors at compile-time, without resorting to panics. Actually emitting the error on the exact token would require dropping the quoting, which would require tighter integration with Lean's parsing machinery. This also adds some tests which identify that the current parser is suboptimal.

### DIFF
--- a/Kraken/Parser.lean
+++ b/Kraken/Parser.lean
@@ -857,7 +857,7 @@ instance {T1} : Coe (ParseResult (List T1) (Sigma String.Pos)) (Except String (L
   | .error _ .eof => .ok []
   | .error _ (.other msg) => .error msg
 
-def parse (input: String): Except String Program := do
+def parse (input: String) : Except String Program := do
   let rawLines := (input.splitOn "\n")
   let (_, lines) ← rawLines.foldlM (fun (lineNum, acc) x => do
     match (parseLine ⟨ x, x.startPos ⟩ : Except String (List Directive)) with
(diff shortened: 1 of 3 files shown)


Kraken/Parser.lean
```diff
@@ -866,10 +866,10 @@ def parse (input: String): Except String Program := do
   ) ((1 : Nat), [])
   pure lines.reverse.flatten
 
-/-- Parse an assembly string, panicking on failure (for use in #eval). -/
-def parse! (input : String) : Program :=
-  match parse input with
-  | .ok prog => prog
-  | .error msg => panic! s!"parse error: {msg}"
+/-- A version of `parse` that runs at compile-time. -/
+elab "parse(" s:str ")" : term => do
```
samuelgruetter, line 870: The changes seem to make sense, except for the naming, for someone coming from Rocq/Bedrock2/Rust: There, any "special" functions like macros or Ltac that run immediately usually end with a `!`, could we keep this convention? Otherwise, it seems that absence or presence of a space after `parse` determines which version of `parse` gets invoked?

andres-erbsen: Eric tells me that C-style function calls without spaces between function and argument are already Notation-only syntax in Lean

andres-erbsen (approved):

Kraken/Parser.lean
eric-wieser: For what it's worth, lean is inconsistent about the use of `!` in identifiers; `opt.get!` and `seq[i]!` mean "this runs at the usual time, but will panic", while the `s!"Hello {where}"` notation is for string formatting. I guess `asm!"mov ..."` would also be a reasonable syntax, but probably this is not worth bikeshedding too much.

PR #81, opened 2026-04-16
fix: report rsp in the state summary

eric-wieser:
Rather than manually converting to JSON, we can ask Lean to do it for us.

It's not clear to me whether in https://github.com/AeneasVerif/kraken/pull/70/changes#r3096719019, the omission of `rsp` was deliberate.

### DIFF
--- a/Kraken/Semantics.lean
+++ b/Kraken/Semantics.lean
@@ -66,7 +66,7 @@ structure Reg64s where
   r13 : UInt64 := 0
   r14 : UInt64 := 0
   r15 : UInt64 := 0
-  deriving Repr, BEq, DecidableEq, Hashable, Hashable, Lean.ToExpr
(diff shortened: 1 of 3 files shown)


miriampolzer:
Thanks for checking these!
Yes, the omission of rsp was deliberate because we don't compare it (kraken will not match with real assembly anyways). No harm in adding it though, the python script also skips rsp for comparison. And if not, we should least add a comment to avoid future confusion. 

I had the custom json converter in there to get easier parsing in python and enhance readability when printing in command line. It was outputting numbers for unit64 and directly converting from pair to map for registers and flags.
Was there any issue you saw with doing the conversion in lean as opposed to python?

eric-wieser:
Happy to close this in that case, just checking if you were aware that `Lean.ToJson` was derivable. Indeed some comments here would be good, but probably not very high priority.

PR #83, opened 2026-04-17
Fix bug parsing imul

aferr:
Fix silent parse failures in `imul` by adding `attempt` for backtracking

The parser for the `imul` instruction was failing to parse valid 2-operand forms (like `imulq %rdx, %rax`) because it lacked backtracking when attempting to parse the 3-operand form first.

In Lean's parser combinators, the alternative operator `p1 <|> p2` only tries `p2` if `p1` fails without consuming any input. Since the 3-operand branch successfully consumed the first operand and the comma before failing to find a third operand, it would not backtrack to try the 2-operand form, causing the whole parse to fail.

This commit:
1. Wraps the 3-operand parsing branches in `attempt` to ensure the parser state is rolled back on failure, allowing correct fallback to the 2-operand form.
2. Adds a new regression test `test_imul_variants.S` that verifies both 2-operand (suffixed and unsuffixed) and 3-operand forms of `imul`.

### DIFF
--- a/Kraken/Semantics.lean
+++ b/Kraken/Semantics.lean
@@ -276,7 +276,7 @@ inductive Operation (w : Width)
   | cmp  (a : RegOrMem w) (b : Operand w)
   | mul  (src : RegOrMem w)
   | mulx (hi lo : Reg w) (src : RegOrMem w) -- {_ : 32 <= w.bits}
-  -- | imul1 (src : RegOrMem w)
+  | imul1 (src : RegOrMem w)
   | imul (_ : Option (Dst w)) (src1 : RegOrMem w) (src2 : Operand w)
   -- Bitwise
   | test (a : RegOrMem w) (b : Operand w)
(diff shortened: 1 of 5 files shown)


aferr:
The test demonstrates the fix (it will pass against this version and fail against main's Parser.lean)

protz (approved):
I think this works. Thanks for the simplification -- @miriampolzer and I fixed a bunch of these missing `attempt` combinators earlier. Obviously there's still some left...

Some questions.
- There is a one-operand form for imul. Do we care?
- ~~I do not understand the first comment in the test. Why did the suffixed two-operand test use to fail, when your fix only concerns the unsuffixed form?~~ my bad it was in the PR description
- It would be nice to have a test that relies on the inference capabilities of the parser, such as e.g. `imul [%rbx] %rax` to make sure things work. Also exercising the three-operand form unsuffixed would be good, e.g. `imul $0, [%rax], %rbx`

aferr:
> There is a one-operand form for imul. Do we care?

I thought may as well support it given that you raised it. I thought using a specialized `imul1` instruction for the semantics for this case is best. (This was commented out in the semantics, so I presume this meant someone else also thought the same?) I checked this against the felixcoultier reference and extended tests to cover this.

> It would be nice to have a test that relies on the inference capabilities of the parser

Done

aferr:
I'll wait for re-review before merging since there is now a new instruction semantics here. But I'm happy with this change if you are so feel free to merge if it looks OK.

Kraken/Semantics.lean
```diff
@@ -455,6 +457,23 @@ def Operation.interp {α} [∀ w : Width, Undefined w.type α] [Undefined Status
...
+  -- imul1 and imul collectively describe variants of the same
+  -- syntax level `imul` instruction, where imul1 is the 1-operand case
+  | .imul1 src =>
+    let a := s.regs.get (Reg.low .rax w)
+    src.interp s p (fun b =>
+    let v := a.toInt * b.toInt
+    let s := if w == .W8 then
+      s.setReg (.low .rax .W16) (BitVec.ofInt 16 v)
+    else
+      let low := BitVec.ofInt w.bits v
```
protz, line 469: @aferr sorry for the delay -- I was at a lean workshop with Andres and it took me a little while to catch up

can you add a comment saying why the bit-manipulations are necessary? I was expecting something like:

```
let result = BitVec.ofInt (w.bits * 2) v
let low = result.take w.bits
let high = result.drop w.bits
```

which might be clearer?

aferr: Sorry that I merged this! I was not sure if anyone would have time to re-review. I can address that in a follow up.

protz: no worries at all :-)

PR #86, opened 2026-04-17
fix sbb subtracting 2^64

andres-erbsen:
UInt64.toInt returns a non-negative number
BitVec.toInt returns may return a negative number
BitVec.toNat returns a Nat which has saturating subtraction

Let's stick with the Rocq convention of .unsigned and .signed for fixed-width variables, both returning Int

h/t @miriampolzer for asking me about this code

### DIFF
--- a/Kraken/Semantics.lean
+++ b/Kraken/Semantics.lean
@@ -9,6 +9,8 @@ attribute [-instance] BitVec.instNatCast
 attribute [-instance] BitVec.instIntCast
 instance : Coe Bool Nat where coe := Bool.toNat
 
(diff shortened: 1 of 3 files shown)


protz:
Saturating Nat subtraction always trips us, so I'm in support of this. What's up with take vs truncate?

andres-erbsen:
They're the same function, just `take` doesn't make me go and look up which end is being truncated. I can put `truncate` back if you prefer it, or we could do something like `truncLsb`.

protz (approved):
I like the consistency. Good for me 👍

Issue #77, opened 2026-04-16
OSX support for test suite


protz:
#70 significantly overhauled the test suite, but support for OSX was not ported over from #65 

This requires investigating the usage of _main (which seemingly creates a full-fledged executable that dynamically links against the system library) vs _start (which is much more barebones on OSX than on Linux).

andres-erbsen:
Can we use Intel SDE (the emulator) as a testing reference?

miriampolzer:
I looked a bit deeper into this; here is my understanding and some proposals on what we can do.

Newer MacBooks are all ARM64. In particular, both the MacBooks on the GitHub CI and Jonathan's work MacBook should be ARM64.
- On GitHub's ARM64 Mac runners, it would use Rosetta 2, which is a translator, not x86 hardware. I am pretty convinced this is also what was used to run https://github.com/AeneasVerif/kraken/pull/65 locally.
- Rosetta 2 implements a deliberate subset of x86-64: no BMI1/BMI2, no ADX, and no AVX (except opt-in in recent macOS versions, and still without full coverage).
- Any test that exercises ADX/BMI2/AVX under Rosetta SIGILLs. `test_arithmetic.S` is the first one to hit this because of `adcxq`/`adoxq`/`mulxq`.
- On top of that, Rosetta's licensing for commercial CI use is ambiguous. We would have to verify permissions to run it before setting up CI runs with Rosetta.

Other issues with running these tests under Rosetta seem to be cleanly solvable by switching to `clang` and fixing some other small issues (syscall differences, register zeroing, etc.). Here is a draft that runs tests successfully under both Linux and OSX: https://github.com/AeneasVerif/kraken/compare/main...miriampolzer:kraken:mac_ci
In particular, I have added a simple list for tests that are to be skipped, and added the arithmetic test.

Regarding the SDE emulator: I don't think it will run on an ARM64 Mac, see the [FAQ here]( 
https://www.intel.com/content/www/us/en/developer/articles/tool/software-development-emulator.html).
> Intel SDE runs on IA-32 or Intel® 64 processors running Windows or Linux or OS X operating systems.

But maybe I am missing something regarding SDE. Let me know.

My suggested path would be:
- Do not run macOS on the CI. It provides no additional value except for verifying that the Mac setup still works. Also, the licensing issue would need to be clarified beforehand (which may not be easy).
[...]

PR #87, opened 2026-04-18
test against Intel SDE

andres-erbsen:
~~May also address https://github.com/AeneasVerif/kraken/issues/77~~

### DIFF
--- a/Kraken/Test/asm_tests.py
+++ b/Kraken/Test/asm_tests.py
@@ -6,12 +6,17 @@ import struct
 import subprocess
 import sys
 import tempfile
+import shutil
+import platform
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
 SCRIPT_DIR = Path(__file__).resolve().parent.parent.parent
 KRAKEN_RUNNER = SCRIPT_DIR / ".lake/build/bin/krakenrunner"
+LOCAL_SDE_RUNNER = SCRIPT_DIR / "tool" / "sde" / "sde64"
+SDE_RUNNER = LOCAL_SDE_RUNNER if LOCAL_SDE_RUNNER.exists() else shutil.which("sde64")
+NATIVE_X64 = platform.machine().lower() in ["x86_64", "amd64"]
 
 REGS = ["rax", "rbx", "rcx", "rdx", "rsi", "rdi", "rsp", "rbp",
         "r8", "r9", "r10", "r11", "r12", "r13", "r14", "r15"]
@@ -25,14 +30,12 @@ class Color:
     BOLD = "\033[1m"
     RESET = "\033[0m"
 
-def get_boilerplate(instruction_text: str) -> str:
-  reg_count = len(REGS)
-  # We move all base registers + the eflags register into memory, so as to dump it later to stdout.
-  total_bytes = (reg_count + 1) * 8
-
-  moves = "\n    ".join([f"movq %{reg}, _final_state + {i*8}(%rip)" for i, reg in enumerate(REGS)])
-
-  return f"""
+def get_boilerplate_native(instruction_text: str) -> str:
+    reg_count = len(REGS)
+    # We move all base registers + the eflags register into memory, so as to dump it later to stdout.
+    total_bytes = (reg_count + 1) * 8
+    moves = "\n    ".join([f"movq %{reg}, _final_state + {i*8}(%rip)" for i, reg in enumerate(REGS)])
+    return f"""
 .data
 .align 8
 _final_state: .space {total_bytes}
@@ -61,6 +64,17 @@ _start:
     syscall
 """
 
+def get_boilerplate_sde(instruction_text: str) -> str:
+    return f"""
+.text
+.globl _start
+_start:
+# --- Test Code Start ---
+{instruction_text}
+# --- Test Code End ---
+    int3
+"""
+
 @dataclass
 class ExecutionState:
     regs: Dict[str, int]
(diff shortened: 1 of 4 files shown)


protz (approved):
This generally looks good to me. Want me to test this on OSX?

Kraken/Test/asm_tests.py
```diff
@@ -6,12 +6,17 @@
...
+import platform
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
 SCRIPT_DIR = Path(__file__).resolve().parent.parent.parent
 KRAKEN_RUNNER = SCRIPT_DIR / ".lake/build/bin/krakenrunner"
+LOCAL_SDE_RUNNER = SCRIPT_DIR / "tool" / "sde" / "sde64"
+SDE_RUNNER = LOCAL_SDE_RUNNER if LOCAL_SDE_RUNNER.exists() else shutil.which("sde64")
+NATIVE_X64 = platform.machine().lower() in ["x86_64", "amd64"]
```
miriampolzer, line 19: I just learned about this, so I may be wrong, but I am surprised the SDE works on anything but native x86. For which configurations did you test this?

Kraken/Test/asm_tests.py
```diff
@@ -61,6 +64,17 @@ def get_boilerplate(instruction_text: str) -> str:
...
 
+def get_boilerplate_sde(instruction_text: str) -> str:
+    return f"""
+.text
+.globl _start
+_start:
+# --- Test Code Start ---
+{instruction_text}
+# --- Test Code End ---
+    ud2
```
miriampolzer: Neat. We should do it in native code too (if we actually keep both).

Kraken/Test/asm_tests.py
```diff
@@ -71,20 +85,43 @@ def parse_raw_state(raw_bytes: bytes) -> ExecutionState:
...
         flags={name: bool(rflags & (1 << bit)) for name, bit in FLAG_MAP.items()}
     )
 
-def run_real_x86(asm_path: Path) -> Tuple[Optional[ExecutionState], Optional[str]]:
+def parse_sde_debugtrace(trace_path: Path) -> ExecutionState:
+    regs = {r: 0 for r in REGS}
+    rflags_val = 0x202
+    with open(trace_path) as f:
+        for line in f:
+            if line == '	XDIS 000000000040103a: BASE 0F0B                     ud2':
```
miriampolzer: The address here might change, right? Maybe future proof: `if "ud2" in line.lower() and "xdis" in line.lower()`

Kraken/Test/asm_tests.py
```diff
@@ -98,6 +135,36 @@ def run_real_x86(asm_path: Path) -> Tuple[Optional[ExecutionState], Optional[str
...
+
+        full_source = get_boilerplate_sde(asm_path.read_text())
+        s_file.write_text(full_source)
+
+        try:
+            subprocess.run(["as", "-o", str(obj_file), str(s_file)], check=True, capture_output=True)
+            subprocess.run(["ld", "-o", str(bin_file), str(obj_file)], check=True, capture_output=True)
+
+            # SDE will exit with 132 (SIGILL) on ud2, so we don't check=True here.
+            subprocess.run(
```
miriampolzer: Not checking for issues here will make debugging real issues with the assembly in a test harder.
Better catch and distinguish between the expected error and other errors. We still lose some errors, but keep e.g. reporting segfaults.
You will need to pull the trace parsing out of the try block, here is roughly how I think this could work:
```
try:
  subprocess.run(["as", "-o", str(obj_file), str(s_file)], check=True, capture_output=True)
  subprocess.run(["ld", "-o", str(bin_file), str(obj_file)], check=True, capture_output=True)
  subprocess.run(
    [str(SDE_RUNNER), "-debugtrace", "-odebugtrace", str(trace_file), "--", str(bin_file)],
    check=True, capture_output=True, timeout=TIMEOUT_SECONDS
  )
            
  except subprocess.CalledProcessError as e:
    cmd_name = Path(e.cmd[0]).name
            
    if cmd_name == SDE_RUNNER.name:
      if e.returncode in (132, -4):
```
128 plus 4, or -4, that is just a guess though, please confirm. maybe also simulate a segfault in the assembly to check that the else case actually catches something.
```
        pass 
      else:
        err_out = (e.stderr or e.stdout or b"").decode(errors="replace").strip()
        return None, f"SDE Crash (Exit Code {e.returncode}):\n{err_out}"
```
If I recall correctly the error handling below wasnt super useful in the runtime crash case, so maye this is better. If not, just fall through to the else case below.
```
    else:
      err = (e.stderr or b"").decode(errors="replace").replace(str(tmp), "...").strip()
      prologue_len = full_source.split("# --- Test Code Start ---")[0].count("\n") + 1
      line_nr_adjusted_err = re.sub(r":(\d+):", lambda m: f":{int(m.group(1)) - prologue_len}:", err)
      return None, f"x86 Error ({cmd_name}):\n{line_nr_adjusted_err}"

if not trace_file.exists():
  return None, "SDE Error: SDE trace file was not generated."
            
  return parse_sde_debugtrace(trace_file), None
```

By the way, if you have a second maybe you can clean up a timeout catch I had not included:
```
except subprocess.TimeoutExpired:
   return None, f"Error: Execution timed out after {TIMEOUT_SECONDS} seconds."
```

miriampolzer (changes requested):

Kraken/Test/asm_tests.py
andres-erbsen: In native code, failing with ud2 would not let us get the state out.

Kraken/Test/asm_tests.py
andres-erbsen: I switched to a different exit condition.

Kraken/Test/asm_tests.py
andres-erbsen: Just testing on my Intel Linux laptop. I hear at least get-sde.sh fails on OSX+ARM.

Kraken/Test/asm_tests.py
andres-erbsen: I found a way to get SDE to stop early without an error! PTAL

Kraken/Test/asm_tests.py
```diff
@@ -71,46 +85,93 @@ def parse_raw_state(raw_bytes: bytes) -> ExecutionState:
...
+        except subprocess.SubprocessError as e:
             err = (e.stderr or b"").decode(errors="replace").replace(str(tmp), "...").strip()
             prologue_len = full_source.split("# --- Test Code Start ---")[0].count("\n") + 1
             line_nr_adjusted_err = re.sub(r":(\d+):", lambda m: f":{int(m.group(1)) - prologue_len}:", err)
-            return None, f"x86 Error ({e.cmd[0]}):\n{line_nr_adjusted_err}"
+            return None, f"x86 Error: {e}\nStderr:\n{line_nr_adjusted_err}"
+
+def run_sde(asm_path: Path) -> Tuple[Optional[ExecutionState], Optional[str]]:
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        tmp = Path(tmp_dir)
```
miriampolzer, line 158: nit: Can you pull out the file path setup and the ld/as commands into helper functions? That way we do not have to adapt in two places when stuff changes.

Kraken/Test/asm_tests.py
```diff
@@ -71,46 +85,93 @@ def parse_raw_state(raw_bytes: bytes) -> ExecutionState:
...
+
+        try:
+            subprocess.run(["as", "-o", str(obj_file), str(s_file)], check=True, capture_output=True)
+            subprocess.run(["ld", "-o", str(bin_file), str(obj_file)], check=True, capture_output=True)
+            subprocess.run([str(SDE_RUNNER), "-stop_int3", "1", "-early_out",
+                            "-dt_lines", "-dt_rawinst", "-dt_print_funcsim",
+                            "-debugtrace", "-odebugtrace", str(trace_file), "--",
+                            str(bin_file)], check=True, capture_output=True, timeout=TIMEOUT_SECONDS)
+        except subprocess.SubprocessError as e:
+            return None, f"SDE Error: {e}\nStderr:\n{(e.stderr or b'').decode(errors='replace').strip()}"
```
miriampolzer, line 169: I think we still want line adjusted errors here, so that as and ld issues are not confusing to debug (you can test this by putting nonsense in an asm file, should give you the right line number to fix).

Kraken/Test/asm_tests.py
```diff
@@ -71,46 +85,93 @@ def parse_raw_state(raw_bytes: bytes) -> ExecutionState:
...
         return ExecutionState(regs=data["regs"], flags=data["flags"]), None
-    except subprocess.CalledProcessError as e:
-        return None, f"Kraken Error:\n{(e.stderr or b"").decode(errors="replace").strip()}"
-    # This except clause ensures any stderr messages are shown even if there is a timeout
-    # (The default Exception object does not have a stderr attribute, so we cannot show this there)
-    except subprocess.TimeoutExpired as e:
-        return None, f"Kraken Error: {e}\nStderr:\n{(e.stderr or b'').decode(errors='replace').strip()}"
-    except Exception as e:
-        return None, f"Kraken Error: {e}"
+    except subprocess.SubprocessError as e:
```
miriampolzer, line 179: nit: It is good that we catch timeouts now, but I think the error is not going to show that this was a timeout, because stderr won't really say. Might be good to add/keep a separate clause for timeout errors (or an if clause to check for timeout and different error printing - up to you).

Kraken/Test/asm_tests.py
```diff
@@ -121,33 +182,39 @@ def get_undefined_flags(path: Path) -> List[str]:
         return [f.strip() for f in raw_flags.split(",") if f.strip()]
     return []
 
-def compare_states(real: ExecutionState, kraken: ExecutionState, undefined_flags: List[str]) -> List[str]:
+def compare_states(real: ExecutionState, kraken: ExecutionState, undefined_flags: List[str], name : str) -> List[str]:
```
miriampolzer: nit: real_type or real_name or similar, to make it clear what this name is.

Kraken/Test/asm_tests.py
miriampolzer, line 238: Fail early with a message if neither native x64 nor sde runner available, otherwise tests will silently pass despite not being compared to anything.

I am not sure if there are valid runs where we are not native x64 but sde runner is available (32 bit architectures, maybe?), so there may be some dead logical paths here and we could just fail if not x64.

miriampolzer (approved):
Looks good apart from some small changes (see comments)! Thanks for adding this.

Kraken/Test/asm_tests.py
miriampolzer: heh, nice :) If I understand correctly that's like a debugging breakpoint and we stop execution when it's reached, so that's exactly what we would want.

Kraken/Test/asm_tests.py
protz: https://www.intel.com/content/www/us/en/download/684897/intel-software-development-emulator.html does not list a download for OSX so indeed we should just not bother with OSX

I'll figure things out

Kraken/Test/asm_tests.py
andres-erbsen: worksforme?

```
$ cat timeout.py 
import subprocess
try:
    subprocess.run(
        ["sh", "-c", "echo 'Analyzing execution state...' >&2; sleep 2"],
        timeout=1, capture_output=True, check=True)
except subprocess.SubprocessError as e:
    print(e, e.stderr)
$ python timeout.py 
Command '['sh', '-c', "echo 'Analyzing execution state...' >&2; sleep 2"]' timed out after 1 seconds b'Analyzing execution state...\n'
```

Kraken/Test/asm_tests.py
andres-erbsen: `reference_name`?

PR #88, opened 2026-04-18
round-trip tests and mov* parser bugfix

andres-erbsen:
To be squashed.

### DIFF
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -19,5 +19,4 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: leanprover/lean-action@v1
-      - run: lake build krakenrunner
       - run: python3 Kraken/Test/asm_tests.py Kraken/Test/asm
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -28,5 +28,5 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: leanprover/lean-action@v1
-      - run: lake build krakenrunner
+      - run: lake build krakenrunner att2intel
       - run: python3 Kraken/Test/asm_tests.py Kraken/Test/asm
--- a/Att2Intel.lean
+++ b/Att2Intel.lean
@@ -1,0 +1,22 @@
+-- Parses AT&T syntax assembly into Kraken's Lean format and then prints it as Intel syntax.
+import Kraken.Parser
+import Kraken.PrintIntel
+
+open Kraken.Parser
+
+def main (args : List String) : IO UInt32 := do
+  if args.isEmpty then
+    IO.eprintln "Usage: att2intel <assembly.S>"
+    return 1
+
+  let asmCode ← IO.FS.readFile args[0]!
+
+  match Kraken.Parser.parse asmCode with
+  | .ok prog =>
+      IO.println ".intel_syntax noprefix"
+      for d in prog do
+        IO.println (toString d)
+      return 0
+  | .error e =>
+      IO.eprintln s!"Parse Error: {e}"
+      return 1
--- a/Att2Intel.lean
+++ b/Att2Intel.lean
@@ -1,0 +1,21 @@
+import Kraken.Semantics
+import Kraken.Parser
+
+open Kraken.Parser
+
+def main (args : List String) : IO UInt32 := do
+  if args.isEmpty then
+    IO.eprintln "Usage: att2intel <assembly.S>"
+    return 1
+
+  let asmCode ← IO.FS.readFile args[0]!
+
+  match Kraken.Parser.parse asmCode with
+  | .ok prog =>
+      IO.println ".intel_syntax noprefix"
+      for d in prog do
+        IO.println (toString d)
+      return 0
+  | .error e =>
+      IO.eprintln s!"Parse Error: {e}"
+      return 1
--- a/Kraken/Parser.lean
+++ b/Kraken/Parser.lean
@@ -621,7 +621,7 @@ def parseInstr : Parser Instr := do
     let w ← instrWidth mn
     commaSeparated w parseOperand parseRegOrMem .mov
 
-  | "moszx" =>
+  | "movsx" =>
     -- Must be a register otherwise lacking type info
     let ⟨ _w_src, src ⟩ ← parseRegW
     let ⟨ w_dst, dst ⟩ ← parseRegW
@@ -639,7 +639,7 @@ def parseInstr : Parser Instr := do
     let w_src ← Char.toWidth c_src
     let src ← parseRegO w_src; parseComma
     let dst ← parseRegO w_dst
-    pure ⟨ .W64, w_dst, .movzx dst (.reg src) ⟩
+    pure ⟨ .W64, w_dst, .movsx dst (.reg src) ⟩
 
   | "movzbw" | "movzbl" | "movzbq" | "movzwl" | "movzwq" =>
     let w_dst ← instrWidth mn
@@ -844,7 +844,7 @@ def parseInstr : Parser Instr := do
       skipHWs
       let sz ← parseHexOrDec
       pure ⟨ .W64, .W64, .nop sz.toNat ⟩
-    ) <|> (pure ⟨ .W64, .W64, .nop 0 ⟩)
+    ) <|> (pure ⟨ .W64, .W64, .nop 1 ⟩)
 
   -- Control flow - conditional jumps
   | _ =>
--- a/Kraken/Parser.lean
+++ b/Kraken/Parser.lean
@@ -541,7 +541,7 @@ def parseInstr : Parser Instr := do
     let w ← instrWidth mn
     commaSeparated w parseOperand parseRegOrMem .mov
 
-  | "moszx" =>
+  | "movsx" =>
     -- Must be a register otherwise lacking type info
     let ⟨ _w_src, src ⟩ ← parseRegW
     let ⟨ w_dst, dst ⟩ ← parseRegW
@@ -559,7 +559,7 @@ def parseInstr : Parser Instr := do
     let w_src ← Char.toWidth c_src
     let src ← parseRegWithType w_src; parseComma
     let dst ← parseRegWithType w_dst
-    pure ⟨ .W64, w_dst, .movzx dst (.Reg src) ⟩
+    pure ⟨ .W64, w_dst, .movsx dst (.Reg src) ⟩
 
   | "movzbw" | "movzbl" | "movzbq" | "movzwl" | "movzwq" =>
     let w_dst ← instrWidth mn
@@ -764,7 +764,7 @@ def parseInstr : Parser Instr := do
       skipHWs
       let sz ← parseHexOrDec
       pure ⟨ .W64, .W64, .nop sz.toNat ⟩
-    ) <|> (pure ⟨ .W64, .W64, .nop 0 ⟩)
+    ) <|> (pure ⟨ .W64, .W64, .nop 1 ⟩)
 
   -- Control flow - conditional jumps
   | _ =>
--- a/Kraken/PrintIntel.lean
+++ b/Kraken/PrintIntel.lean
@@ -1,5 +1,9 @@
 import Kraken.Syntax
-
+/-!
+# IntelPrinter
+This file prints Kraken-supported x64 assembly in Intel syntax.
+This is different from Parser.lean which expects AT&T syntax.
+-/
 
 instance : ToString Reg64 where
   toString r := match r with
--- a/Kraken/PrintIntel.lean
+++ b/Kraken/PrintIntel.lean
@@ -1,0 +1,136 @@
+import Kraken.Syntax
+
+
+instance : ToString Reg64 where
+  toString r := match r with
+  | .rax => "rax" | .rbx => "rbx" | .rcx => "rcx" | .rdx => "rdx"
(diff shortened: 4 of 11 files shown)


Kraken/Semantics.lean
```diff
@@ -799,3 +801,143 @@ namespace Reg
...
+instance : ToString RegOrRip where
+  toString
+  | .ofRegW ⟨_, r⟩ => ToString.toString r
+  | .rip => "rip"
+
+def ConstExpr.toStr : ConstExpr → String
+  | .Label l => l
+  | .Int64 i => if i < 0 then s!"-{i.abs}" else s!"{i}"
+  | .before_current_instruction => "."
+  | .after_current_instruction => "."
```
andres-erbsen: I don't know how to print these; currently only the latter is used

andres-erbsen: We don't seem to be parsing it either, unless I hear otherwise I will remove this constructor

.github/workflows/lean_action_ci.yml
```diff
@@ -28,5 +28,5 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: leanprover/lean-action@v1
-      - run: lake build krakenrunner
+      - run: lake build krakenrunner att2intel
```
miriampolzer: Since krakenrunner is now in the default build, we can remove this line. To catch build errors as early as possible, we should include all targets (including att2intel) in the default build. This keeps the CI configuration clean while still allowing users to build targets individually if they choose.

Kraken/Test/asm_tests.py
```diff
@@ -6,12 +6,14 @@
...
 import sys
 import tempfile
+import difflib
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
 SCRIPT_DIR = Path(__file__).resolve().parent.parent.parent
 KRAKEN_RUNNER = SCRIPT_DIR / ".lake/build/bin/krakenrunner"
+ATT2INTEL = SCRIPT_DIR / ".lake/build/bin/att2intel"
```
miriampolzer: nit: Let's make this `BIN_DIR = Path(__file__).resolve().parent.parent.parent / ".lake/build/bin"`.
SCRIPT_DIR was poor naming from my side, and it isn't used anywhere else so we can ditch it.

Kraken/Test/asm_tests.py
```diff
@@ -129,8 +130,66 @@ def compare_states(real: ExecutionState, kraken: ExecutionState, undefined_flags
             diffs.append(f"flag {f}: x86={real.flags[f]} | kraken={kraken.flags[f]}")
     return diffs
 
+
+def assembled_text(src_path: Path, tmp_dir: Path) -> Tuple[Path, bytes]:
+    obj_path = tmp_dir / f"{src_path.stem}.o"
+    bin_path = tmp_dir / f"{src_path.stem}.bin"
+    subprocess.run(['as', "-c", "-o", str(obj_path), str(src_path)],
```
miriampolzer: What is the `-c` flag for here? I just checked on my machine and `as` doesn't seem to have it.

Kraken/Test/asm_tests.py
```diff
@@ -129,8 +130,66 @@ def compare_states(real: ExecutionState, kraken: ExecutionState, undefined_flags
...
+                   stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+    subprocess.run(["objcopy", "-O", "binary", "-j", ".text", str(obj_path), str(bin_path)],
+                   stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
```
miriampolzer: I think you can use `capture_output=True` instead of setting `PIPE` for both `stdout` and `stderr`.

Kraken/Test/asm_tests.py
```diff
@@ -129,8 +130,66 @@ def compare_states(real: ExecutionState, kraken: ExecutionState, undefined_flags
...
+    return obj_path, bin_path.read_bytes()
+
+def disassemble(obj_path: Path) -> List[str]:
+    res = subprocess.run(["objdump", "-d", str(obj_path)],
```
miriampolzer: nit: Catch and print errors for all subprocesse, here and for the assembled_text function above. I suppose these should never happen in normal test runs but makes debugging changes easier.

It's fine to still crash everything, I think:
```
except subprocess.CalledProcessError as e:
        print(f"Subprocess failed: {e.cmd}")
        print(f"Error: {e.stderr.decode()}")
        raise
```

Kraken/Test/asm_tests.py
```diff
@@ -129,8 +130,66 @@ def compare_states(real: ExecutionState, kraken: ExecutionState, undefined_flags
...
+    colored = []
+    for line in diff_lines:
+        if line.startswith('---') or line.startswith('+++'):
+            colored.append(f"{Color.BOLD}{line}{Color.RESET}")
+        elif line.startswith('-'):
+            colored.append(f"{Color.RED}{line}{Color.RESET}")
+        elif line.startswith('+'):
+            colored.append(f"{Color.GREEN}{line}{Color.RESET}")
+        elif line.startswith('@@'):
+            colored.append(f"\033[96m{line}{Color.RESET}")
```
miriampolzer: nit: Add to colors (CYAN?).

Kraken/Test/asm_tests.py
```diff
@@ -129,8 +130,66 @@ def compare_states(real: ExecutionState, kraken: ExecutionState, undefined_flags
...
+        except subprocess.CalledProcessError as e:
+            err_msg = e.stderr.decode().strip() if isinstance(e.stderr, bytes) else (e.stderr or "")
+            return False, f"att2intel failed: {' '.join(str(x) for x in e.cmd)}\n{err_msg}"
+
+        try:
+            orig_obj, orig_bytes = assembled_text(asm_path, tmp_dir)
+            intel_obj, intel_bytes = assembled_text(intel_src, tmp_dir)
+            if orig_bytes != intel_bytes:
+                diff = difflib.unified_diff(
+                    disassemble(orig_obj), disassemble(intel_obj),
```
miriampolzer, line 197: This is the part where I have little experience so I may be asking something obvious.

If you have a difference in the assembled files and then disassemble both to see the diff, is that usually going to give you a meaningful hint as to what's wrong?

Are there cases where the assembled bytes may differ, but not in a meaningful way? I.e. cases where these tests would throw false failures, like alignment, different encodings for the same thing, nop instructions...
If that can happen, maybe only fail if both the binary and the disassembled texts differ?

Kraken/Test/asm_tests.py
```diff
@@ -145,6 +204,11 @@ def test_file(path: Path) -> Tuple[bool, str]:
         print(f"[{Color.RED}FAIL{Color.RESET}]")
         return False, "\n".join(diffs)
```
miriampolzer, line 226: I’d say we can turn these around and run the new tests first.

If the translation logic is off, the execution is pretty much doomed to fail anyway, but with way less helpful output. Catching a bug through an objdump diff is should be a lot faster than trying to figure out why a register state is wrong by tracing the translation through semantics.

If tests are nice and small and self-contained then the new tests you are adding should give a much better debugging output (if they fail) than the result diff.

Att2Intel.lean
```diff
@@ -0,0 +1,21 @@
+import Kraken.Semantics
+import Kraken.Parser
+
+open Kraken.Parser
+
```
miriampolzer, line 6: Can you add a one-line comment what this does? E.g. "Parses AT&T syntax assembly into Kraken's Lean format and the prints it as Intel syntax.".

Kraken/Semantics.lean
```diff
@@ -799,3 +801,131 @@ namespace Reg
 @[match_pattern] abbrev r14b := low .r14 .W8
 @[match_pattern] abbrev r15b := low .r15 .W8
 end Reg
+
```
miriampolzer, line 634: Would it make sense to put this in a separate PrintIntel.lean file?

I checked everything here for naming and it looks right to me, as for the lean syntax I can't give much meaningful input.

miriampolzer (commented):
Cool! That looks useful, thanks for adding. My comments are mostly details as to how exactly we want to run these, and some cleanups.

I think we need good naming to make it easy to grasp what the two differnt tests do. Maybe something along the lines of (Static) Translation Parity (binary comparison - what you are adding) and (Dynamic) Semantic Fidelity (execution result diffs - the exisitng tests)?

Kraken/Semantics.lean
samuelgruetter: Yes, please put the printing code into a separate file 👍 This will hopefully also help avoid merge conflicts with #100, where I split Semantics into separate files for Syntax and Semantics.

Kraken/Test/asm_tests.py
andres-erbsen: Yes, both could happen. The latter relatively rare for x86, though I'm aware of cases where it is intentionally triggered (e.g. padding with redundant prefixes instead of extra instructions).

Kraken/PrintIntel.lean
```diff
@@ -0,0 +1,136 @@
+import Kraken.Syntax
+
+
```
aferr, line 7: Maybe add a short comment at the top explaining that this gives toStrings for the intel syntax of x86 (by contrast to ATT)?

Kraken/Parser.lean
```diff
@@ -639,7 +639,7 @@ def parseInstr : Parser Instr := do
     let w_src ← Char.toWidth c_src
     let src ← parseRegO w_src; parseComma
     let dst ← parseRegO w_dst
-    pure ⟨ .W64, w_dst, .movzx dst (.reg src) ⟩
+    pure ⟨ .W64, w_dst, .movsx dst (.reg src) ⟩
```
aferr, line 642: Thanks for fixing this! No need to change this now but for future PRs could you mention any bug fixes in the PR description (espeically if it is a side effect to the original point of the PR).

Att2Intel.lean
```diff
@@ -0,0 +1,22 @@
+-- Parses AT&T syntax assembly into Kraken's Lean format and then prints it as Intel syntax.
```
aferr, line 1: Just out of curiosity: what is the motivation for this? Is it to catch bugs in our parser?

.github/workflows/lean_action_ci.yml
```diff
@@ -19,5 +19,4 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: leanprover/lean-action@v1
-      - run: lake build krakenrunner
```
aferr, line 22: Why is this deleted?

Kraken/Test/asm_tests.py
```diff
@@ -133,8 +135,82 @@ def compare_states(real: ExecutionState, kraken: ExecutionState, undefined_flags
...
+                    fromfile="AT&T", tofile="Intel", lineterm=""
+                )
+                colored = colorize_diff(diff)
+                return False, f"Roundtrip mismatch:\n" + "\n".join(colored)
+        except subprocess.CalledProcessError as e:
+            err_msg = e.stderr.decode().strip() if isinstance(e.stderr, bytes) else (e.stderr or "")
+            return False, f"assembler failed: {' '.join(str(x) for x in e.cmd)}\n{err_msg}"
+    return True, ""
+
 def test_file(path: Path) -> Tuple[bool, str]:
```
aferr, line 207: Did you try running this, for example on some of the assembly files we already have for the test runner? Is it much work to include a few round trip tests on assembly files in CI?

aferr (approved):
My comments are optional. Looks good including bug fix.

Att2Intel.lean
andres-erbsen: Yes, this is for testing our parser.

.github/workflows/lean_action_ci.yml
andres-erbsen: It's redundant with https://github.com/AeneasVerif/kraken/blob/main/lakefile.toml#L3

Kraken/Test/asm_tests.py
andres-erbsen: I think it is running on the CI for this PR? [log here](https://productionresultssa7.blob.core.windows.net/actions-results/38fe22ef-d21d-49bd-a43e-259aa083a88c/workflow-job-run-56af41cb-2824-56c6-8984-2afa08c17d0f/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-07-01T16%3A14%3A26Z&sig=HHrG%2BYIFcL1RkhisPKLCJPGpoKHFsF6nhZqepE53tf4%3D&ske=2026-07-01T19%3A49%3A59Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-07-01T15%3A49%3A59Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-07-01T16%3A04%3A21Z&sv=2025-11-05) where I think "PASS" is printed by [this line](https://github.com/AeneasVerif/kraken/pull/88/changes/BASE..06b28a83df2d3efeb7620780081048f50d519d2f#diff-982eef09577277c9caf34bf12a9ac01536b9d42149df4e066f6761fee808bba2R228).

Kraken/PrintIntel.lean
andres-erbsen: https://github.com/AeneasVerif/kraken/pull/88/changes/beeb741b17f05f2001936cc4f7284ba943a256bb

Att2Intel.lean
aferr: thanks!

PR #89, opened 2026-04-18
fix pop instruction to correctly handle %rsp as destination

andres-erbsen:


protz (approved):

protz (changes requested):
https://www.felixcloutier.com/x86/pop says "Loads the value from the top of the stack to the location specified with the destination operand (or explicit opcode) and **then** increments the stack pointer."

Your patch seems to do the opposite. Am I missing something?

andres-erbsen:
Indeed. But the test fails without this change and passes with it (even with SDE from #87).

Kraken/Test/asm/test_pop_mem_rsp.S
```diff
@@ -0,0 +1,15 @@
+mov %rsp, %rsi
+
+pushq %rsp
+pushq %rsp
+popq (%rsp)
```
andres-erbsen, line 5: Even more fun: what does this one do?

andres-erbsen:
After thinking about it some more, I suspect the sentence in the documentation exists to order steps 1 and 2 below, and is just incorrect when read as ordering steps 2 and 4.
1. load-address calculation may use rsp
2. rsp is incremented
3. destination-address calculation may use rsp
4. the destination is written with the loaded value (not rsp)

It would of course be great if we could find an authoritative reference to confirm this behavior.

protz:
As long as Intel SDE agrees I think we're good? Thanks for investigating.

But I'm still confused because the pseudocode says:

```
    ELSE IF StackAddrSize = 64
        THEN
                IF OperandSize = 64
                    THEN
                        DEST := SS:RSP; (* Copy quadword *)
                        RSP := RSP + 8;
```

let's talk about it today

protz:
@lukaszobernig mentioned this https://github.com/NationalSecurityAgency/ghidra/issues/4282

protz (approved):

PR #93, opened 2026-04-21
try to abstract over memory permissions & MMIO without abstracting over state

samuelgruetter:
As discussed one-on-one with @andres-erbsen, here's an attempt to abstract over memory access behavior without abstracting over the state, as an alternative to #60 and #72.

The intended typical usage (not yet coded up) would be to split the system's state into two parts: user-code-visible `MachineState` as modeled currently, and use-case-specific OS/device state. A use-case-specific top-level "one-step" function would match on the instruction and intercept it if it is a syscall that modifies OS state, or else dispatch to `Instr.interp`.

The `Permissions` typeclass could be used e.g. to mark some pages as read-only, e.g. to model read-only memory that stores e.g. constant tables for cryptographic functions, or it could be used to implement memory protection keys that limit what pages can be accessed based on a (yet-to-model) PKRU register and per-page 4-bit protection keys.

The `NonmemAccess` typeclass could be used to model all the addresses that don't behave like regular memory, such as MMIO.

I haven't tried using it yet, but I think the `Permissions` abstraction could (roughly) work.

However, I don't see how to get `NonmemAccess` access to work: `nonmem_store` (and potentially also `nonmem_load`) might modify state that is not part of `MachineState`, but "system state" such as the state of an external device, but the current type signatures do not allow this. In an imperative language, we could simply pass in closures that mutate the system state, but in a functional setting, I don't (yet?) see how to allow this without abstracting over the state.

### DIFF
--- a/Kraken/Examples.lean
+++ b/Kraken/Examples.lean
@@ -29,7 +29,6 @@ example [layout : Layout] s : step1 (layout p1) (s, layout.start) (fun s => s.1.
   dsimp only [Directives.interp,Directive.interp,Instr.interp,Operation.interp,Operand.interp]
   dsimp only [MachineData.set,Reg64s.set,MachineData.setReg,Reg64s.set64,ConstExpr.interp]
   simp (ground:=True)
-  simp
 
   /- simp [Instr.interp,Operation.interp,Operand.interp,MachineData.set] -/
   /- simp [MachineData.setReg,Reg64s.set,Reg64s.set64,ConstExpr.interp] -/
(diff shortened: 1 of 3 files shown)


andres-erbsen:
In bedrock2 we had I/O traces to model nonmem access, and for modeling alone the same could make sense here. However, the desire to also compose with an interpreter for these I/O actions seems thoroughly reasonable. I think that's what itree work is built for. idk if that's the wisest choice for us, but "just enough itree" could be worth a try.

PR #95, opened 2026-04-23
Make krakenrunner a default target.

aferr:
I'm not sure about anyone else, but I often go:
```
lake build
... attempt to run tests...
... Oh no!! tests broken.. why?
... Oh... the krakenrunner is stale ;(
```
This change just adds krakenrunner to the default targets to avoid this mistake.

### DIFF
--- a/lakefile.toml
+++ b/lakefile.toml
@@ -1,6 +1,6 @@
 name = "Kraken"
 version = "0.1.0"
-defaultTargets = ["Kraken"]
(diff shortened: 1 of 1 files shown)


aferr:
Totally fine to not do this if others don't want it for some reason

protz:
that's great thanks

protz:
does this mean we can simplify the CI definitions to no longer build krakenrunner explicitly?

miriampolzer:
Yes, that should work.

PR #96, opened 2026-04-24
Improve readability of imul1 semantics

aferr:
Implement suggestion [here](https://github.com/AeneasVerif/kraken/pull/83#discussion_r3137994946) which I agree improves readability.


protz (approved):
Thanks!

PR #97, opened 2026-04-24
effectful semantics

andres-erbsen:
Alternative to https://github.com/AeneasVerif/kraken/pull/93

@samuelgruetter wdyt?

### DIFF
--- a/Kraken/Examples.lean
+++ b/Kraken/Examples.lean
@@ -26,9 +26,10 @@ example [layout : Layout] s : step1 (layout p1) (s, layout.start) (fun s => s.1.
   dsimp only [step1,Executable.straightline]
   rw [Executable.directivesFromStart]
   simp [List.mapIdx,List.mapIdx.go]
-  dsimp only [Directives.interp,Directive.interp,Instr.interp,Operation.interp,Operand.interp]
-  dsimp only [MachineData.set,Reg64s.set,MachineData.setReg,Reg64s.set64,ConstExpr.interp]
+  dsimp only [Directives.interp,Directive.interp,Instr.interp,Operation.interp,Operand.interp,RegOrMem.interp]
+  dsimp only [MachineData.set,Reg64s.set,MachineData.setReg,Reg64s.set64,ConstExpr.interp,Sem.all]
   simp (ground:=True)
+  simp
 
   /- simp [Instr.interp,Operation.interp,Operand.interp,MachineData.set] -/
   /- simp [MachineData.setReg,Reg64s.set,Reg64s.set64,ConstExpr.interp] -/
@@ -54,7 +55,7 @@ theorem swap_correct [layout : Layout] (d : MachineData) :
   simp [List.mapIdx, List.mapIdx.go]
   -- TODO It would be nice to progress instruction by instruction instead of all at once, like below.
   dsimp only [Directives.interp, Directive.interp, Instr.interp, Operation.interp, Operand.interp, RegOrMem.interp]
-  dsimp [MachineData.set, MachineData.setReg, Reg64s.set, Reg64s.set64]
+  dsimp [MachineData.set, MachineData.setReg, Reg64s.set, Reg64s.set64, Sem.all]
   intros _af1 _af2 _af3
   apply Eventually.done
   simp (ground:=True)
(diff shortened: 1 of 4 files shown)


samuelgruetter:
Looks exciting!

It seems that the `α` now became the abstract state I was wishing for -- as in, I can instantiate it with a struct that contains the `MachineData` plus whatever additional external-device state I want!

I wrote a dummy MMIO-controlled device [here](https://github.com/samuelgruetter/kraken/commit/efa12e517b491bb0ec3b2389af2de23ed58089d6).

I still want to test it, and to understand what I actually did, so far I just did trial-and-error-until-it-typechecks, guided by some high-level intuition...

andres-erbsen:
PTAL -- I think it's better not to put system state around Sem rather than inside it.

samuelgruetter:
It seems that `Sem` does not actually need a type parameter α:

https://github.com/samuelgruetter/kraken/commit/833f4bea0aaba2870c2b28fb8af0f2a8e49d4e3e

I also added an IncrementDMAExample on this branch:

https://github.com/samuelgruetter/kraken/commits/mmio-on-top-of-effectful-semantics/

It illustrates all cells in the 2x2 matrix `{ nonmem_store, nonmem_load } x { DeviceState, MachineState }`. Each of the 4 cells corresponds to a case of "effect might modify state" that I would like to support. To achieve this, it seems that `Sem` needs to store a snapshot of the `MachineState` not only in the `.ret` constructor, but also in all `nonmem_load` and `nonmem_store` constructors, and if we want to keep hashing like we did before [this change](https://github.com/AeneasVerif/kraken/commit/acfecf0867fbf5e7bf651557cec9bc535d56f24e#diff-8dbfd56672d93618866485b5ca26aca396a303057c4203c2efb968eb53f64473R114-R120), we also need to add a snapshot of the `MachineState` to the `undefined_` constructors. Do you think this is the way to go @andres-erbsen?

andres-erbsen:
- If we take away \alpha from .ret, maybe we should rename it? I lack ideas though. `.next_instruction`?
- Yes, nonmem_load/store should be allowed to transfer memory. Let's have the semantics reveal memory to the effect handler, and expect memory back from it. But let's not reveal CPU registers or flags.
- For undefined, let's just reveal the hash of the machine state.
- You *can* make handle_effects direct-style, but a CPS definition is more convenient for direct proofs (and you can define non-CPS by passing `id` to the CPS one, but not vice versa)

andres-erbsen:
Jonathan and I discussed this and we would like to continue pursuing effectful semantics in this style.

samuelgruetter:
> * If we take away \alpha from .ret, maybe we should rename it? I lack ideas though. `.next_instruction`?
> * Yes, nonmem_load/store should be allowed to transfer memory. Let's have the semantics reveal memory to the effect handler, and expect memory back from it. But let's not reveal CPU registers or flags.
> * For undefined, let's just reveal the hash of the machine state.
> * You _can_ make handle_effects direct-style, but a CPS definition is more convenient for direct proofs (and you can define non-CPS by passing `id` to the CPS one, but not vice versa)

I agree with all of these and addressed them [on my branch](https://github.com/samuelgruetter/kraken/commits/mmio-on-top-of-effectful-semantics/). I also added some other tweaks.

samuelgruetter:
> Jonathan and I discussed this and we would like to continue pursuing effectful semantics in this style.

Awesome! 

I have a bunch of use cases in mind that such effectful semantics should support, and I'm trying them out on my branch.

samuelgruetter:
My current "blocker" is that I would like to record unsupported instructions as effects, so that effect handlers can add custom handling for them, but this requires that `Effect` (formerly `Sem`) is defined *after* `Operation` and `Instr`. This raises the question of whether you want to separate syntax and semantics more strictly? Currently, the Parser imports all of Semantics, which is not so modular. But having the definitions of X and X.interp close to each other is also nice. Any thoughts on how to proceed?

andres-erbsen:
I think it's fine to define Effect after Instr, and we probably will want Syntax in a separate file anyway.

andres-erbsen:
re `Effect` vs `Sem` -- I've seen Effect used for the parameter of the `interact` case of itrees/freer monads, and have not seen it used for the entire type of shallowly embedded interactive programs. We could choose to use it this way, but `Sem` as the domain of X64 semantics would kinda makes sense if it was X64.Sem (and we ought to rename that module...)

samuelgruetter:
> re `Effect` vs `Sem` -- I've seen Effect used for the parameter of the `interact` case of itrees/freer monads, and have not seen it used for the entire type of shallowly embedded interactive programs.

Interestingly, in [this Haskell package](https://hackage.haskell.org/package/freer-effects-0.3.0.1/docs/Control-Monad-Freer.html), they define `data Eff effs a`, i.e. they use plural to denote the type parameter for *one* effect event (from a union of many possible effects), and singular to denote the entire type of shallowly embedded interactive programs -- exactly the opposite of my intuition :shrug:

PR #100, opened 2026-04-28
Effectful semantics & small use-case demos

samuelgruetter:
Here's a rebase of https://github.com/AeneasVerif/kraken/pull/97 with additional small use-case demos to check if the type signatures make sense.

Current demos include:
- an MMIO scenario with an interpreter
- a description of an MMIO+DMA scenario where MMIO reads and writes both can affect both internal and external state
- a custom ISA extension, obtained by implementing an effect handler for the `unsupported_instruction` effect


samuelgruetter:
I updated the PR as follows:

* As suggested by @andres-erbsen, I changed `Operand.interp` to take and return "too much" state, namely the full `MachineData`. I believe that it will never modify the registers, and only rarely (namely in case of MMIO loads changing memory ownership) modify the memory. However, the advantage of always taking and returning the full `MachineData` is that the state-threading becomes very lightweight: We just need an additional `s` argument in each continuation, which shadows the previous `s`. So the resulting code in `Operation.interp` is now probably quite close to what we'd get by unfolding the monad-related definitions of a [CPS state monad](https://github.com/samuelgruetter/kraken/blob/11f17605a60d5f06282fab5ded6a8d72e0656366/Kraken/Examples/CPSSMachineStateMonad.lean).
* I removed the `unsupported_instruction` effect, lenient parsing mode and the `CustomInstructions` example because they might be controversial in the current form. The changes are preserved on my [effectful-use-cases-v2](https://github.com/samuelgruetter/kraken/commits/effectful-use-cases-v2/) branch and I will bring back an improved, hopefully less controversial version in a separate PR.
* Minor tweaks like making `BitVec.replace` more readable
* Renaming of the effects and changing the permissions signatures to combine the access test and the handle-permission-failure into one effect

Should be ready for (the next) review now!

Kraken/Examples/IncrementMMIO.lean
```diff
@@ -0,0 +1,124 @@
...
+-/
+
+-- state as seen by software, not necessarily implemented like this in hardware
+inductive IncrementerState
+  | idle
+  | busy (input : UInt32) (steps_until_done : Nat)
+  | done (answer : UInt32)
+  deriving Hashable
+
+inductive Incrementer.Register | VALUE | STATUS
```
eric-wieser, line 36: Note that making these all uppercase goes against the lean naming conventions

Kraken/Examples/IncrementMMIO.lean
```diff
@@ -0,0 +1,124 @@
...
+  | .VALUE => match s with
+    | .idle => some (.busy v N_STEPS_NEEDED)
+    | .busy _ _ => none
+    | .done _ => none
+  | .STATUS => none
+
+def IncrementerState.internal_step (s : IncrementerState) : IncrementerState :=
+  match s with
+  | .idle => s
+  | .busy input n => if n == 0 then .done (input + 1) else .busy input (n - 1)
```
eric-wieser: nit: could also write as
```suggestion
  | .busy input 0 => .done (input + 1)
  | .busy input (n + 1) => .busy input n
```

Kraken/Examples/IncrementMMIO.lean
```diff
@@ -0,0 +1,124 @@
...
+  else if addr == STATUS_ADDR then some .STATUS
+  else none
+
+structure SystemState where
+  machineState : MachineState
+  deviceState : IncrementerState
+
+def handle_effects (ds : IncrementerState) (es : Effects)
+  (ok : SystemState → Except String SystemState)
+: Except String SystemState :=
```
eric-wieser: Lean style:

```suggestion
def handleEffects (ds : IncrementerState) (es : Effects)
    (ok : SystemState → Except String SystemState) :
    Except String SystemState :=
```

Kraken/Examples/IncrementMMIO.lean
```diff
@@ -0,0 +1,124 @@
...
+  | .require_exec_access _ cont => handle_effects ds (cont ()) ok
+  | .nonmem_load dmem addr w cont =>
+    match w with
+      | .W32 => match Incrementer.Register.of_addr (UInt64.ofBitVec addr) with
+        | .some r => match ds.read_step r with
+          | .some (reply, ds') =>
+            handle_effects ds' (cont (UInt32.toBitVec reply) dmem) ok
+          | .none => .error s!"Incrementer.read_step failed"
+        | .none => .error s!"nonmem_load at unmapped address {repr addr}"
+      | _ => .error s!"nonmem_load of width other than 4 bytes"
```
eric-wieser: This is quite a bit easier to read as
```suggestion
  | .nonmem_load dmem addr w cont => do
    let .W32 := w
      | throw s!"nonmem_load of width other than 4 bytes"
    let some r := Incrementer.Register.of_addr (UInt64.ofBitVec addr)
      | throw s!"nonmem_load at unmapped address {repr addr}"
    let some (reply, ds') := ds.read_step r
      | throw s!"Incrementer.read_step failed"
    handle_effects ds' (cont (UInt32.toBitVec reply) dmem) ok
```

Kraken/Examples/IncrementDMA.lean
```diff
@@ -0,0 +1,68 @@
...
+
+inductive IncrementerState.write_step
+  (s : IncrementerState) (r : Incrementer.Register) (v : UInt64)
+  (s' : IncrementerState) : Prop where -- TODO
+
+inductive IncrementerState.internal_step
+  (s : IncrementerState)
+  (s': IncrementerState) : Prop where -- TODO
+
+def Incrementer.Register.of_addr (addr : UInt64) : Option Incrementer.Register :=
```
eric-wieser, line 57: naming convention:
```suggestion
def Incrementer.Register.ofAddr (addr : UInt64) : Option Incrementer.Register :=
```

Kraken/Examples/IncrementMMIO.lean
```diff
@@ -0,0 +1,124 @@
...
+  match r with
+  | .VALUE => match s with
+    | .idle => none
+    | .busy _ _ => none
+    -- note how a *read* causes a state transition in the device
+    | .done answer => some (answer, .idle)
+  | .STATUS => match s with
+    | .idle => some (STATUS_IDLE, s)
+    | .busy _ _  => some (STATUS_BUSY, s)
+    | .done _ => some (STATUS_DONE, s)
```
protz, line 56: I assume it is intentional here that there should be no state transition, i.e. the device can't move onto the next step unless the value has been read out?

Kraken/Examples/IncrementMMIO.lean
```diff
@@ -0,0 +1,124 @@
...
+    | .done _ => none
+  | .STATUS => none
+
+def IncrementerState.internal_step (s : IncrementerState) : IncrementerState :=
+  match s with
+  | .idle => s
+  | .busy input n => if n == 0 then .done (input + 1) else .busy input (n - 1)
+  | .done _ => s
+
+def Incrementer.Register.of_addr (addr : UInt64) : Option Incrementer.Register :=
```
protz, line 79: is this where you defined the memory mapping?

Kraken/Examples/IncrementMMIO.lean
```diff
@@ -0,0 +1,124 @@
...
+-- software and proofs should not depend on this number,
+-- if we have nondeterminism available, we'd pick an arbitrary number here
+def N_STEPS_NEEDED : Nat := 3
+
+def IncrementerState.write_step (s : IncrementerState)
+  (r : Incrementer.Register) (v : UInt32)
+  : Option IncrementerState :=
+  match r with
+  | .VALUE => match s with
+    | .idle => some (.busy v N_STEPS_NEEDED)
```
protz, line 67: can you say more about the N_STEPS_NEEDED? is this akin to some fuel, or can this be stated in terms of `eventually` for more generality about some unspecified number of steps before the device is ready?

Kraken/Semantics.lean
```diff
@@ -108,50 +87,85 @@ structure StatusFlags where
...
+  else require_read_access addr w (fun _unit =>
+    let key := UInt64.ofBitVec (addr &&& ~~~0b111#64)
+    match s.dmem[key]? with
+    | .some v => ret (v.toBitVec.extractLsb' ((addr &&& 0b111#64) * 8#64).toNat w.bits) s
+    | .none => nonmem_load s.dmem addr w (fun v dmem => ret v { s with dmem }))
+
+def MachineData.store (s : MachineData) (addr : BitVec 64) {w : Width} (v : w.type) (ret: MachineData → Effects) : Effects :=
+  if addr % w.bytesv != 0 then .unimplemented s!"Unimplemented: only aligned memory access is supported"
+  else require_write_access addr w (fun _unit =>
+    let key := UInt64.ofBitVec (addr &&& ~~~0b111#64)
```
protz, line 159: why the mask when we assert above that the lower bits are already zero?

Kraken/Semantics.lean
```diff
@@ -698,31 +594,31 @@ def Executable.directivesFromAddress (e : Executable) (a : Int64) : List (Direct
...
-  | .error s => .error s
-partial_fixpoint
+partial def Executable.eval (e : Executable) (s : MachineState) (until_ : MachineState → Bool) : Except String (MachineState) :=
+  if until_ s then .ok s else handle_effects (e.straightline s .done)
+where
+  handle_effects es :=
+    match es with
+    | .done s => eval e s until_
+    | .unimplemented msg => .error msg
+    | .require_read_access _ _ ok => handle_effects (ok ())
```
protz: is this going to leave `handle_effects` at the head of our goal?

protz (commented):
I looked at this quickly and so far it looks good. I've only added minor comments.

Two high-level questions that I don't have an answer to, but maybe you do :-).
1. How is this going to impact the shape of our goals? Are we always going to have `handle_effects` in the front of our goal, and does this mean our tactics will have to evolve to reduce underneath `handle_effects`?
2. I do not see yet how the transfer of ownership from MMIO'd device back to main CPU is going to work. Right now, in `nonmem_load`, I only see a memory being returned to the continuation -- should there be room for an additional predicate that could express e.g. a permission over a certain fraction of the memory that the device has now relinquished?

Thanks!

Kraken/Examples/IncrementMMIO.lean
samuelgruetter: This is a deliberate choice. In hardware reference manuals, the names of memory-mapped device registers typically (or I might even claim always, having read a [lot](https://arxiv.org/abs/2510.27485) of hardware reference manuals recently) have uppercase names. And I also use uppercase names for constants that would appear in the C code of a driver for such a device.

Kraken/Examples/IncrementMMIO.lean
samuelgruetter: Exactly! The device holds the value until it was read. And, remarkably, *reading* the value, which might sound effect-free, does have the side effect of putting the device back into idle state.

Kraken/Examples/IncrementMMIO.lean
samuelgruetter: Exactly!

Kraken/Examples/IncrementMMIO.lean
samuelgruetter: I want to make sure both simulation (running an evaluator inside Lean on an example with concrete values) and proofs work. In IncrementMMIO, I want to test simulation, and in IncrementDMA, I want to test proofs. So, here in the simulator example, I chose to use fuel, but in the proof example, I will try to allow the device to nondeterministically decide at each step whether it's done or not. However, this can cause problems to prove termination: A driver that busy-waits for to a device that might happen to decide to always be not yet done would never terminate. So maybe we'll need some kind of fuel in the proof example as well -- at least that's the solution that seemed to work in [silveroak](https://github.com/project-oak/silveroak/blob/cccfdb4e19c5906256d2ed4487bf1a0aa2fab99e/firmware/IncrementWait/IncrementWaitSemantics.v#L78-L83).

And yes, driver proofs will use `eventually`! And maybe we can get rid of the fuel by *assuming* an `eventually` for the device?

Kraken/Semantics.lean
samuelgruetter: because the mask always zeros the 3 least significant bits, whereas the modulo check might look at the 0, 1, 2, or 3 least significant bits

Kraken/Semantics.lean
samuelgruetter: `simp` can get rid of it, but I did have to update some proofs that were passing explicit allow-lists to `simp` or `dsimp`

samuelgruetter:
> 1. How is this going to impact the shape of our goals? Are we always going to have `handle_effects` in the front of our goal, and does this mean our tactics will have to evolve to reduce underneath `handle_effects`?

First, a naming nitpick: I used `handle_effects` for simulators (designed for `#eval`), and `Effects.all` for proofs. I would not recommend writing proofs about simulators, so I'm answering a modified version of your question where `handle_effects` is replaced by `Effects.all` in `Tactics.lean`.

Before this PR, there were already effects, namely `[Undefined T α]` and `[Throw α]`, modeled as an individual typeclass for each effect. Proofs had to unfold the definition of the `throw` and `undefined` projections of [these instances](https://github.com/AeneasVerif/kraken/blob/main/Kraken/Tactics.lean#L19-L23). Now, with this PR, they just need to unfold `Effects.all` instead, so I think not much will change.

One difference, though, is that we can now have chains (trees) of effects, but proofs should never consider trees of more than one instruction, and one instruction will not have too many effects, so it should be fine to unfold all of `Effects.all` at once.

Another difference is the addition of the `require_read/write/exec` permissions. They are an additional permission check on top of checking whether the address is in `MachineData.dmem`. Most use cases can instantiate the handlers for these effects to a no-op. Other use cases like reasoning about MPK or read-only pages will instantiate them with something more interesting and will therefore have to perform a little additional reasoning step at each memory access.

> 2. I do not see yet how the transfer of ownership from MMIO'd device back to main CPU is going to work. Right now, in `nonmem_load`, I only see a memory being returned to the continuation -- should there be room for an additional predicate that could express e.g. a permission over a certain fraction of the memory that the device has now relinquished?

I intend `require_read/write/exec_access` to be orthogonal to memory ownership protocols between drivers and devices. The former is intended to model the restriction that the OS/MMU/MPK enforce, whereas the latter is to enforce restrictions that come from English text in hardware reference manuals, like e.g. "once the NIC has advanced the head pointer of this circular queue of buffer descriptors, it takes ownership of the buffers and software may not modify them anymore until it gets an interrupt or reads a done flag that indicates that the NIC is done filling this buffer".

Now, to model memory ownership changes that result from protocols between drivers and devices, the plan is that software giving up ownership is modeled as removing the memory from `MachineData.dmem`, and putting it into the external state of the device. Once the device gives back memory to software, it is removed from the device's state, and put back into `MachineData.dmem`. This does not directly allow modeling memory that software may only write, but hardware may write, but I think that's fine, because I'd like to keep the invariant that everything in  `MachineData.dmem` is exclusively owned by software. To model memory (e.g. done flags) that hardware may write and software may read, I'd use the same approach as for memory-mapped registers.

All of this is an approach to define operational semantics. It does not preclude inventing a more advanced separation logic on top of this.

I did not (yet) document or illustrate all the details of MMIO+DMA because I believe that Andres and I are very much on the same page about this. We developed these ideas during our PhD, but both of us graduated before we got a chance to publish them. So, if you have in-person access to Andres, that's the best source to clarify :wink:

samuelgruetter:
Thanks everyone for the feedback!

I think I've addressed all of it.

I also checked that `lake build && lake build krakenrunner && python3 Kraken/Test/asm_tests.py Kraken/Test/asm` succeeds on my machine, but it would also be good to see if it succeeds on CI, which would require that someone with write access clicks a button to confirm that I'm not trying to mine bitcoin on your CI :wink:

Kraken/Examples/IncrementMMIO.lean
eric-wieser: Fair enough, perhaps worth a remark to that effect in the module docstring.

protz (approved):
Andres and I talked about this some more -- it all makes sense for me now. We are quite excited about having support for MMIO/DMA'd devices and we think merging sooner rather than later is better, especially as we are about to write more tactics.

Thanks!

PR #102, opened 2026-04-28
Prove missing lemma from Examples

protz:
Not intending to merge this immediately, but I felt slightly annoyed that there was this blatant sorry in Examples.

Turns out, with the proper lemmas in scope (which are available in 4.29 but not earlier) about `scanl`, this went throught automatically.

Let's merge once we feel comfortable upgrading the toolchain. At least now we know the lemma is true :-)


protz:
superseded by #107

Issue #101, opened 2026-04-28
Change eventually to rely on `step1` rather than `straightline`


protz:
As reported by @aferr:

`Eventually` as defined in `Tactics.lean` currently relies on `step1` that directly defers to `straightline`.

Instead, `step1` should represent a single step of execution, with an auxiliary lemma in the style of `eventually_trans`, stating that if one reaches `p` from `initial` by running `straightline`, then `eventually program p initial`. Or something along those lines :-)

samuelgruetter:
I agree! Moreover, instead of abstracting over `(prog: Executable)`, `eventually` should abstract over the state type, and over a step predicate of type `(step: State -> (State -> Prop) -> Prop)`. This `step` predicate can then be instantiated with a partially applied single-instruction-step that remembers prog, and different use cases can also pick different variants of a single-instruction step.

I think of `eventually` as an operator that lifts a single step to multiple steps, in the same way as the transitive closure operator (`star`) lifts a single step in traditional small-step operational semantics to multiple steps. See e.g. the section "Omni-Smallstep Semantics" of [this blog post](https://samuelgruetter.net/blog/2022/09/30/omnisemantics/#:~:text=We%20can%20also%20define) for a side-by-side comparison of `star` and `eventually`, in case Andres hasn't already preached this to you 😉 

It might be best to define `eventually` (and maybe also `always`) in a separate file, similar to how [we did in Rocq](https://github.com/mit-plv/coqutil/blob/cfd88cef005cda5ec400108606d27a0d40578a8f/src/coqutil/Semantics/OmniSmallstepCombinators.v).

PR #104, opened 2026-05-01
Parameterize Omnisemantics for Different Transitions

aferr:
This is essenitally to implement #101 . A concrete instantiation in terms of Executable.straightline remains, a new one in terms of Executable.step is now given. A theorem relating the former to the latter is stated with progress towards proof, though may not be completed within this change.


Kraken/Tactics.lean
```diff
@@ -22,78 +22,89 @@ instance : Throw Prop where
...
-      (mid_p: Post) ->
-      step1 prog initial mid_p →
-      (forall (mid: MachineState), mid_p mid → Eventually _ p mid) →
-      Eventually _ p initial
-
--- THEOREMS
-
-theorem step_cps {p : Executable} [Layout] (post: Post) (initial: MachineState):
-  step1 p initial (fun mid => Eventually p post mid) → Eventually p post initial :=
+inductive Eventually {State : Type} (trans : State → (State → Prop) → Prop) (post : State → Prop) : State → Prop
```
samuelgruetter: In Rocq, it pays off to put as many arguments as possible to the left of the final `:`, because that results in better elimination/induction principles, as in, if you match on an Eventually, you don't learn anything new about the shape of its argument `initial`, whereas the current signature suggests that you might. In Rocq, putting `initial` to the left of the colon worked, so why not also do it like this in Lean? I see that it was already different from Rocq before this PR, but maybe now is a good time to reconsider or document why it's as-is?

Kraken/Examples.lean
```diff
@@ -20,10 +20,38 @@ theorem Executable.directivesFromStart [layout : Layout] prog :
...
 
+-- Collects all directives at address 'a' by batching all consecutive zero-sized directives (labels) plus
+-- the first non-zero-sized directive. (Concretely, labels are of size zero, for example).
+-- General but awkward in proofs due to takeWhile, dropWhile
+theorem Executable.directivesAtAddress_split [Layout] (e : Executable) (a : Int64) :
+  e.directivesAtAddress a =
+    let ds := e.directivesFromAddress a
+    let zero_sized := ds.takeWhile (fun x => x.2 = 0)
+    let first_nonzero := ds.dropWhile (fun x => x.2 = 0)
+    zero_sized ++ first_nonzero.take 1 := by
```
eric-wieser: Isn't this false via `List.takeWhile_append_dropWhile`?

Kraken/Tactics.lean
eric-wieser: You can't move `initial` here because of `(forall (mid: State), mid_p mid → Eventually trans post mid)` below

aferr: Added a comment to clarify. Thanks both.

Kraken/Theorems.lean
```diff
@@ -23,3 +24,109 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+              obtain ⟨h_mid2, h_interp⟩ := h_and
+              have h_eq_next : ds' = p.directivesFromAddress mid.2 := by
+                rw [h_mid2]
+                exact (directivesFromAddress_tail p pc d sz ds' h_eq.symm).symm
+              apply ih mid.1 mid.2 h_eq_next
+              exact h_interp
+  apply h_gen (p.directivesFromAddress initial.2) initial.1 initial.2 (by rfl)
+  exact h
+
+theorem eventually_straightline_to_step1 [Layout] (p: Executable) (initial: MachineState) (post: Post)
```
aferr: It's not likely I'll be able to prioritize completing this proof anytime soon. Is it still useful to check in this WIP version (where at least one theorem defined is not true) too perhaps in case someone has time to finish it later? Or is the preference not to have partial proofs in the repo? (There's just a few unfinished proofs in examples).

Kraken/Tactics.lean
samuelgruetter: Ok, so I just learnt about a Rocq/Lean difference then!

Kraken/Examples.lean
aferr: Deleting this code in any case. Thanks for giving this your attention! :)

Kraken/Theorems.lean
aferr: Deleted this and left just the top level def since the addition of effects changed this noticably and this isn't really needed yet.

protz (commented):
This looks good! Thanks.

I have  few minor comments.
- It looks like the `Post` abbreviation (`State -> Prop`) is no longer used -- is this intentional?
- With some reshuffling of arguemnts, `step` could become `Post -> Post` which might make things easier? 

@andres-erbsen?

Kraken/Tactics.lean
```diff
@@ -25,78 +25,91 @@ def Effects.all (s : Effects) (post : MachineState → Prop) : Prop :=
...
 
--- TACTIC MACROS
+-- ============================================================================
+-- Concrete Transition Rules
+-- ============================================================================
+
+def step1 [Layout] (p: Executable) (s: MachineState) (post: Post) :=
+  (Executable.step p s .done).all post
+
+def straightline_step [Layout] (p: Executable) (s: MachineState) (post: Post) :=
```
eric-wieser: nit: usually good practice to be explicit about the type
```suggestion
def straightlineStep [Layout] (p: Executable) (s: MachineState) (post: Post) : Bool :=
```
(also, no `_`s in `def` names in Lean)

Kraken/Theorems.lean
```diff
@@ -23,3 +24,12 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
   have hzero : (0 : Nat) % 2^64 = 0 := Nat.zero_mod (2^64)
   rw [h2, hkmod, hzero] at h1
   exact h_ne h1
+
+-- ============================================================================
+-- Omnisemantics Step Mixing (And soundness of straightline instantiation)
+-- ============================================================================
```
eric-wieser: Note that Lean has a native block comment format, `/-! -/`, which results in headings being rendered in documentation.

aferr:
@protz 
> * It looks like the `Post` abbreviation (`State -> Prop`) is no longer used -- is this intentional?

This got lost when the states in the defs became parametric. Thanks for flagging this. This is back now (and I think is a nice version of it given the constraints around implicits).

> * With some reshuffling of arguemnts, `step` could become `Post -> Post` which might make things easier?

Correct me if I misunderstand, but I think your proposal is something like this:
```
def step_fn {State} (trans : State → Post → Prop) : Post → Post :=
  fun (Q : Post) (s : State) => ∃ mid_p, trans s mid_p ∧ ∀ mid, mid_p mid → Q mid
 
 ...
| step (initial: State):
    step_fn trans (Eventually trans post) initial →
    Eventually trans post initial
...
```

A nice thing about this reformulation is that it becomes closer to a weakest precondition / predicate transformer form. A drawback is that we lose the explicit midpiont state and might need to do `obtain` to get it back in induction proofs.

To try to get the best of both worlds, in this iteration I am proposing a predicate transformer view layered on top so one can use either style without changing the fundamental def. See the `Predicate Transformer View` section. Or the main two meaningful pieces here:
```
def step_fn {State : Type} (trans : State → @Post State → Prop) (Q : @Post State) (s : State) : Prop :=
  ∃ mid_p, trans s mid_p ∧ ∀ mid, mid_p mid → Q mid

-- Induction principle in predicate-transformer style:
-- Eventually is the *least* Post satisfying the fixpoint equation.
theorem eventually_ind_pt {State : Type} (trans : State → Post → Prop) (post inv : Post)
    (h_done : ∀ s, post s → inv s)
    (h_step : ∀ s, step_fn trans inv s → inv s) :
    ∀ s, Eventually trans post s → inv s := by
  intro s he
  induction he with
  | done _ hp              => exact h_done _ hp
  | step _ mid_p ht hc ih  => exact h_step _ ⟨mid_p, ht, ih⟩
  ```

Kraken/Theorems.lean
aferr: I am using these banner style comments elsewhere 🙈 so I would rather not change this (at least not yet) sorry

Kraken/Tactics.lean
aferr: Thanks. fixed

aferr:
@andres-erbsen Can you take a look at this? Thank you.

aferr:
Out of band @andres-erbsen suggested I delete stepFn and related along with the new unproven theorem. I also just changed the return type of Eventually to Post. Not sure why I thought a more complicated thing was desired!

andres-erbsen (approved):

PR #106, opened 2026-05-01
Remove exclusion of I/O from README.md

andres-erbsen:


protz (approved):

protz:
Should we also at some point make sure this is presented as a joint FRO/Google/Inria project in the README?

PR #107, opened 2026-05-01
Playing around with SymM

protz:
I experimented a bunch with SymM... the final example is `silly4`, and demonstrates fine-grained reduction in the goal while intro-ing binders rather than inlining them away.

The example is `of_bits`, a recursive function that is intended to capture how our goals look like, with a continuation at type `UInt64 -> Prop`.

Most of the time was spent figuring out plumbing, how to hook this up to regular tactic mode so that I can see what the tactic does, and doing API discovery. I have a tactic that repeatedly reduces the head, step by step, while introducing the binders.

One limitation is that I could not trigger simplifications in the goal. Not sure why.

Comments welcome. I intend to switch to reducing goals from `Semantics.lean` later today, hopefully.


aferr:
Is this draft meant to help with the register state explosion issue for long straightline sections? Or is #78 still the better place to look for that? (I'm keen to dogfood whichever is meant to help with this issue, or both if both are related)

protz:
@aferr @andres-erbsen as discussed this morning:
- ported all examples in Examples.lean to use `kstep`
- showcase how to do things on small examples (please ask if something's unclear)
- demonstrate usage of memory lemmas
- write out more proof steps for the larger example (commented out at the bottom of the file) that show a general strategy for going through the program

A lot of automation could be improved (I think the `intros` should be part of `kstep`) and I think `kstep` could be improved to try to decide (either with decide, bv_decide or grind) conditionals whenever possible, to unblock automatic progress, but that's a start.

aferr:
Thanks. Before this gets merged, would you mind factoring the tactic itself out of Examples.lean and into a separate file? (Perhaps the tactic can go into a new Kstep.lean / Step.lean, and the theorems can go into Theorems.lean. Perhaps it is also time to rename the current Tactics.lean to Omnisemantics.lean).

protz:
This is done, thanks for the feedback. Let me know if this looks good

protz:
Discussed with @andres-erbsen -- merging

andres-erbsen (approved):

PR #108, opened 2026-05-03
Experimental beta let reduction sym simproc

Kha:


protz:
Thanks @Kha. I've been working on SymM this week, and this will be very useful.

Do I understand correctly that SymM is still missing some features, since I see that your prototype reverts to the regular `tactic` mode in a few spots?

PR #109, opened 2026-05-05
We're one rfl away from a CPS state monad

samuelgruetter:
I implemented the `.add` case of `Operation.interp` in CPS-monadic style and showed that it is definitionally equivalent to the current definition. In fact, it's even alpha-equivalent, after unfolding `bind` and `pure` and some argument-order-swappers.

In case some monad fans ever ask why we chose the current style, let's collect our design rationales here:

Advantages of current style:
- no need to unfold monad noise
- names of binders seem better preserved (in the example, `b` is preserved, whereas using monadic style, it gets renamed into `a_1`)

Advantages of using a CPS state monad (illustrated in this PR):
- "nicer" interpreter code

### DIFF
--- a/Kraken/DesignDecisions/CpsStateMonad.lean
+++ b/Kraken/DesignDecisions/CpsStateMonad.lean
@@ -1,0 +1,66 @@
+import Kraken.Semantics
+
+def CpsState (σ α : Type) : Type 1 :=
+  σ → (α → σ → Effects) → Effects
+
+instance {σ} : Monad (CpsState σ) where
+  pure x := fun s k => k x s
+  bind m f := fun s k => m s (fun a s' => f a s' k)
+
+-- primed definitions to change argument order:
+
+def RegOrMem.interp' {w} [Labels] [AddressSize]
+  (o : RegOrMem w) (p : Std.Rco Int64) : CpsState MachineData w.type :=
+  fun s (ret : w.type → MachineData → Effects) => RegOrMem.interp o s p ret
+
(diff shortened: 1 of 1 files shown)


protz:
FYI, I had a little sample file that highlighted the difference between the monadic and CPS styles.

https://gist.github.com/protz/b2816478e1c25e582e9b1c5f52f2dd9a

This was early on and we didn't have effects, so Andres was able to use ExceptCpsT as our monad instance.

Kraken/DesignDecisions/CpsStateMonad.lean
```diff
@@ -0,0 +1,66 @@
...
+
+-- use different arguments so that equality doesn't hold and we can inspect the
+-- result of unfolding without dsimp overeagerly applying rfl
+example [Labels] [address_size : AddressSize] {w : Width}
+  (dst1 dst2 : Dst w) (src1 src2 : Operand w) p s next ignored_jmp
+  (e1: dst1 = dst2) (e2: src1 = src2) :
+  add.interp' dst1 src1 p s next = (Operation.add dst2 src2).interp p s (next ()) ignored_jmp
+:= by
+  dsimp only [add.interp', Operation.interp]
+  dsimp only [bind, pure]
```
eric-wieser, line 54: I would strongly recommend against using `dsimp` in this way; it's better to define explicit `CpsState.mk` and `CpsState.run` casting functions, and write `dsimp` lemmas that simplify `run (x >>= f)` etc, as lean itself does.

PR #113, opened 2026-05-06
Parsing entrypoint for files by contrast to just strings

aferr:
Also included -- preprocessing that skips some assembler directives that Kraken does not support.


protz (approved):
I tried this locally and it's very helpful. @andres-erbsen this ignores data directives; I guess at some point we might want to parse that

PR #115, opened 2026-05-14
Fix address size parsing and inference

### LINKED ISSUE #90: parsing should infer address size similarly to operand size
The following test currently fails because there is not enough information in the semantics AST to truncate the address correctly:

```
movl $0x80000000, %eax
leaq (%eax), %rbx
leaq (%eax,%eax), %rbx
leaq (%eax,%eax,1), %rbx
leaq (%eax,%eax,2), %rcx
leaq (%eax,%eax,4), %rdx
leaq (%eax,%eax,8), %rsi
leaq (%rax), %r8
leaq (%rax,%rax), %r9
leaq (%rax,%rax,1), %r10
leaq (%rax,%rax,2), %r11
leaq (%rax,%rax,4), %r12
leaq (%rax,%rax,8), %r13
```

TBD whether registers in AddrExpr should carry sizes at all (Reg vs Reg64).

protz:
Fixes https://github.com/AeneasVerif/kraken/issues/90

### DIFF
--- a/Kraken/Parser.lean
+++ b/Kraken/Parser.lean
@@ -17,6 +17,71 @@ open Std.Internal.Parsec
 open Std.Internal.Parsec.String
 
 -- ============================================================================
+-- Data structures for type inference
+-- ============================================================================
+
+-- We need to eagerly parse (to move through the syntax), but we may need to
+-- defer choosing a width for those operands that are untyped (as in: may have
+-- any width).
+def MaybeTyped (T: Width → Type) :=
+  Σ (w: Option Width), match w with | .some w => T w | .none => { w: Width } → T w
+
+-- Most of our parsing functions return a MaybeAddrWidth × MaybeTyped T, in case
+-- the operand we just parsed happens to be a memory operand, which thus allows
+-- the assembler to infer the address width used for the instruction (see
+-- comments in Syntax).
+abbrev MaybeAddrWidth := Option Width
+
+-- This could be Option.mergeM if it existed.
+--
+-- No x86 assembly instructions take two memory operands: this means that we
+-- have at most one address width hint per instruction.
+def mergeAddrWidths (w1 w2: MaybeAddrWidth): Parser MaybeAddrWidth :=
+  match w1, w2 with
+  | .none, .none => pure .none
+  | .some w1, .none
+  | .none, .some w1 => pure (.some w1)
+  | .some _, .some _ => fail "can't have two memory operands"
+
+-- If the context provides a type annotation, this forces a given width, or
+-- errors out if an incompatible one has been inferred.
+def ascribe {T: Width → Type} (w: Width) (v: MaybeTyped T): Parser (T w) := do
+  match v with
+  | ⟨ .none, v ⟩ => pure v
+  | ⟨ .some w2, v ⟩ =>
+    if h: w = w2 then
+      pure (h ▸ v)
+    else
+      fail s!"type error: {w} != {w2}"
+
+def ascribeOrInfer {T1 T2} (op1: MaybeAddrWidth × MaybeTyped T1) (next: Parser (MaybeAddrWidth × MaybeTyped T2)): Parser (MaybeAddrWidth × Σ w, T1 w × T2 w) := do
+  let (addr_w2, op2) ← next
+  let (addr_w1, op1) := op1
+  let addr_w ← mergeAddrWidths addr_w1 addr_w2
+  match op1 with
+  | ⟨ .some w1, op1 ⟩ =>
+    let op2 ← ascribe w1 op2
+    pure (addr_w, ⟨ w1, op1, op2 ⟩)
+  | ⟨ .none, op1 ⟩ =>
+    match op2 with
+    | ⟨ .some w2, op2 ⟩ =>
+      pure (addr_w, ⟨ w2, op1, op2 ⟩)
+    | ⟨ .none, _ ⟩ =>
+      fail "missing type annotation"
+
+-- Naming convention: O = function takes an operand width
+def parseO {T1} (op: Parser (MaybeTyped T1)) (w: Width): Parser (T1 w) := do
+  let op ← op
+  ascribe w op
+
+-- Naming convention: A = function also returns an address width
+def parseAO {T1} (op: Parser (MaybeAddrWidth × MaybeTyped T1)) (w: Width): Parser (MaybeAddrWidth × T1 w) := do
+  let (addr_w, op) ← op
+  let op ← ascribe w op
+  pure (addr_w, op)
+
+
+-- ============================================================================
 -- Lexical Utilities
 -- ============================================================================
 
@@ -126,16 +191,16 @@ def parseRegW : Parser RegW := do
   let _ ← pchar '%'
   parseRegNameW
 
-def parseRegNameNoHighByte: Parser (Width × Reg64) := do
+def parseLowRegName: Parser (Width × Reg64) := do
   let ⟨ w, r ⟩ ← parseRegNameW
   match r with
   | .low r64 _ => pure (w, r64)
   | _ => fail s!"high byte register cannot be used for an addrexpr"
 
-def parseRegNoHighByte : Parser (Width × Reg64) := do
+def parseLowReg : Parser (Width × Reg64) := do
   skipHWs
   let _ ← pchar '%'
-  parseRegNameNoHighByte
+  parseLowRegName
 
 def parseRegOrRipW : Parser (Width × RegOrRip) := do
   skipHWs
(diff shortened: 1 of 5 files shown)


Kraken/Parser.lean
```diff
@@ -126,16 +126,27 @@ def parseRegW : Parser RegW := do
...
+  let _ ← pchar '%'
+  parseRegNameNoHighByte
+
+def parseRegOrRipW : Parser (Width × RegOrRip) := do
   skipHWs
   let _ ← pchar '%'
   (do
     let _ ← pstring "rip"
-    pure .rip
+    pure (.W64, .rip)
```
protz, line 214: we could parse eip here -- can easily be added later

Kraken/Parser.lean
```diff
@@ -329,28 +357,48 @@ def ascribe {T: Width → Type} (w: Width) (v: MaybeTyped T): Parser (T w) := do
     else
       fail s!"type error: {w} != {w2}"
 
-def ascribeOrInfer {T1 T2} (op1: MaybeTyped T1) (next: Parser (MaybeTyped T2)): Parser (Σ w, T1 w × T2 w) := do
-  let op2 ← next
+def composeAddrWidths (w1 w2: MaybeAddrWidth): Parser MaybeAddrWidth :=
```
protz: maybe there is something better in the Lean standard library?

protz:
I think the renamings are poor, hopefully tomorrow will bring more inspiration

Kraken/Syntax.lean
```diff
@@ -45,11 +45,30 @@ attribute [coe] ConstExpr.int64
...
   deriving Repr, DecidableEq, Hashable, Lean.ToExpr, Hashable, Lean.ToExpr
 
-inductive RegOrRip where | ofRegW (_ : RegW) | rip : RegOrRip
+-- For address expressions, the width of the register is determined only by the
+-- instruction-wide address prefix. Concretely:
+-- * movb $2 (%eax) denotes moving a single byte (operand width) into the memory
+--   location at the 32-bit address stored in %rax (address width) -- note that
+--   there is inference here!
+-- * addr32 movb $2 (%rax) means the exact same thing, so the `addr32` prefix
+--   acts as an instruction-wide override
```
andres-erbsen: I'm not sure it's a good thing that clang accepts this form, but indeed apparently it does. gcc says

```
<source>: Assembler messages:
<source>:1: Error: `(%rax)' is not a valid base/index expression
Compiler returned: 1
```

Kraken/Parser.lean
```diff
@@ -126,16 +126,27 @@ def parseRegW : Parser RegW := do
   let _ ← pchar '%'
   parseRegNameW
 
-def parseRegOrRipW : Parser RegOrRip := do
+def parseRegNameNoHighByte: Parser (Width × Reg64) := do
+  let ⟨ w, r ⟩ ← parseRegNameW
+  match r with
+  | .low r64 _ => pure (w, r64)
+  | _ => fail s!"high byte register cannot be used for an addrexpr"
```
andres-erbsen, line 198: `parseRegName` ?

Kraken/Parser.lean
```diff
@@ -126,16 +126,27 @@ def parseRegW : Parser RegW := do
...
+
+def parseRegNoHighByte : Parser (Width × Reg64) := do
+  skipHWs
+  let _ ← pchar '%'
+  parseRegNameNoHighByte
```
andres-erbsen: `parseRegLow`?

Kraken/Syntax.lean
protz: added a comment to that effect

protz:
Thanks for the good suggestions. I took your comments into account and:
- moved the parsing and type inference-related utilities to their own section
- improved the naming convention to distinguish the four combinations of "takes an expected operand width or not" * "returns a possible address width or not".

This is ready for review.

Kraken/Test/asm/test_lea_addr32.S
```diff
@@ -0,0 +1,13 @@
+movl $0x80000000, %eax
+leaq (%eax), %rbx
```
miriampolzer, line 2: The two values overriden in rbx are not compared in the test, right? Can we use another place to store so that values from line 2 and 3 will be tested? I am thinking r14/15 should work.

Kraken/Parser.lean
```diff
@@ -16,6 +16,71 @@ namespace Kraken.Parser
...
 open Std.Internal.Parsec.String
 
+-- ============================================================================
+-- Data structures for type inference
+-- ============================================================================
+
+-- We need to eagerly parse (to move through the syntax), but we may need to
+-- defer choosing a width for those operands that are untyped (as in: may have
+-- any width).
+def MaybeTyped (T: Width → Type) :=
```
miriampolzer: nit: I think you inherited this but when reading through this the first time I found "Typed" unclear. Maybe you could change it to be called "OpWidth" everywhere? That also clearly shows that addrWidth and opWidth are two separated widths we need to deal with.

Kraken/Parser.lean
```diff
@@ -210,8 +293,16 @@ def parseMemory : Parser AddrExpr := do
   let _ ← pchar ')'
   -- Some adapters between the parsed components and the expected dependent
   -- pairs:
-  let idx := Option.map (fun idx => ⟨idx, scale⟩) idx
-  pure ({ base, idx, disp })
+  let w ← match w1, idx with
+    | w1, .some (w2, _) =>
+      if w1 ≠ w2 then
+        fail "type mismatch in memory addressing operands: base and index have different widths"
```
miriampolzer: nit: print the widths

Kraken/ParserTests.lean
```diff
@@ -147,13 +143,23 @@ info: [Directive.instr
...
+    { address_size := Width.W32, operation_size := Width.W64,
+      operation :=
+        Operation.lea (low Reg64.rax Width.W64)
+          { base := some (RegOrRip.reg Reg64.rbp), idx := some { reg := Reg64.rcx, scale := Width.W32 },
+            disp := ↑16 } }] : List Directive
+-/
+#guard_msgs in
+#check parse("leaq 16(%ebp, %ecx, 4), %rax")
+
 section error_reporting
```
miriampolzer, line 163: nit: Consider adding tests to verify some of the failures regarding width mismatching.

Kraken/Syntax.lean
```diff
@@ -45,11 +45,30 @@ attribute [coe] ConstExpr.int64
...
+-- instruction-wide address prefix. Concretely:
+-- * movb $2 (%eax) denotes moving a single byte (operand width) into the memory
+--   location at the 32-bit address stored in %eax (address width) -- note that
+--   there is inference here!
+-- * addr32 movb $2 (%rax) means the exact same thing, so the `addr32` prefix
+--   acts as an instruction-wide override (at least for Clang; GCC rejects this)
+--
+-- Note that by doing so, we faithfully model two key things
+-- * high byte registers cannot appear within memory operands:
+--
```
miriampolzer: nit: unintential empty line?

Kraken/Semantics.lean
```diff
@@ -176,11 +179,11 @@ def ConstExpr.interp [Labels] : ConstExpr → Std.Rco _root_.Int64 → _root_.In
 
 def AddrExpr.interp [Labels] [address_size : AddressSize] (a : AddrExpr) (s : Reg64s) (p : Std.Rco Int64) :=
   let base := match a.base with
-              | .some (.ofRegW ⟨_, r⟩)  => (s.get r).signed
+              | .some (.reg r) => (s.get64 r).toAddressSize.signed
               | .some .rip => p.upper.toInt
```
miriampolzer, line 183: Do we need to truncate rip too, or is this fine?

Kraken/Test/asm/test_lea_addr32.S
protz: done

Kraken/Parser.lean
protz: done

Kraken/Parser.lean
protz: done

Kraken/ParserTests.lean
protz: done

Kraken/Syntax.lean
protz: fixed comment formatting

Kraken/Semantics.lean
protz: for now we only parse %rip and not %eip -- I've added a comment in the parser saying that the behavior needs to be double-checked once we add support for %eip in AddrExpr

miriampolzer (approved):

PR #116, opened 2026-05-20
Enable parser errors and consecutive labels

miriampolzer:
Propagate parser errors to output instead of silently swallowing them. Parse multiple labels in one line correctly.

### DIFF
--- a/Kraken/Parser.lean
+++ b/Kraken/Parser.lean
@@ -885,47 +885,58 @@ def parseLabelDecl : Parser Label := do
 -- Line and Program Parsing
 -- ============================================================================
 
+def parseAlign : Parser Instr := do
+  let _ ← pstring ".align"
+  skipHWs
+  let alignment ← parseHexOrDec
+  let pad ← (do
+    skipHWs
+    let _ ← parseComma
+    skipHWs
+    let pad ← parseHexOrDec
+    pure (some pad.toNat)
+  ) <|> pure none
+  pure ⟨ .W64, .W64, .nopalign alignment.toNat pad ⟩
+
+def skipSpaceAndCheckLineEnd : Parser Bool := do
+  skipHWs
+  let c? ← peek?
+  match c? with
+  | none
+  | some '\n' =>
+    pure true
+  | some '#' =>
+    skipLineComment
+    pure true
+  | some _ =>
+    pure false
+
+def parseOptionalInstr : Parser (Option Directive) := do
+  if (← skipSpaceAndCheckLineEnd) then
+    pure none
+  else
+    let i ← parseAlign <|> parseInstr
+    pure (some (Directive.instr i))
+
+def checkLineEnd : Parser Unit := do
+  if (← skipSpaceAndCheckLineEnd) then
+    pure ()
+  else
+    fail "unexpected trailing characters on line"
+
 /-- Parse a single line: optional label, followed by optional instruction or directive.
     Returns a list of directives found on the line. -/
 def parseLine : Parser (List Directive) := do
   skipHWs
-  let c ← peek!
-  -- Skip empty lines and comment-only lines
-  if c == '\n' || c == '#' then
-    if c == '#' then skipLineComment
-    pure []
-  else
-    let label ← (attempt do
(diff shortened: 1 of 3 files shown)


Kraken/ParserTests.lean
```diff
@@ -180,25 +180,36 @@ error: line 1: type mismatch in memory addressing operands: base ({w1}) and inde
...
-/-! Behavior which is wrong. -/
-section broken
+/-- error: line 1: unexpected end of input -/
+#guard_msgs in
+#check parse("addq %rax")
 
-/-- info: [] : List Directive -/
+/-- error: line 1: unexpected end of input -/
 #guard_msgs in
 #check parse("xorq %rax, 1")
```
protz, line 189: I'm confused about this example: I think it should parse, and I don't think we should require a newline at the end of the file

Kraken/ParserTests.lean
```diff
@@ -180,25 +180,36 @@ error: line 1: type mismatch in memory addressing operands: base ({w1}) and inde
...
-/-- info: [] : List Directive -/
+/-- error: line 1: unexpected end of input -/
 #guard_msgs in
 #check parse("addq")
 
-/-- error: line 2: unsupported instruction: loop -/
+/-- error: line 2: unexpected end of input -/
 #guard_msgs in
 #check parse("
-loop: loop: addq %rax, %rbx
```
protz, line 199: one thing that we should probably reject: redefining existing labels, but that can be dealt with in a followup pr

Kraken/Test/asm/test_double_label.S
```diff
@@ -0,0 +1,3 @@
+# Undefined flags: cf, pf, af, zf, sf, of
+loop: loop:
```
protz, line 2: is the assembler on CI allowing this? mine says 

```
test.S:1:7: error: symbol 'loop' is already defined
```

Kraken/Parser.lean
```diff
@@ -885,47 +885,58 @@ def parseLabelDecl : Parser Label := do
...
-            let _ ← parseComma
-            skipHWs
-            let pad ← parseHexOrDec
-            pure (some pad.toNat)
-          ) <|> pure none
-          pure (some (Directive.instr ⟨ .W64, .W64, .nopalign alignment.toNat pad ⟩))
-        ) <|> (do
-          let instr ← parseInstr
-          pure (some (Directive.instr instr)))
-    ) <|> pure none
```
protz, line 923: ok I guess this is where parsing was previously allowed to afil

protz (approved):
Looks great, and much cleaner. Thanks! 

I have one question about something that doesn't quite parse, and I'm confused as to why (perhaps it's your change about `.eof` that makes it fail when it shouldn't?).

Once this bit is fixed, let's go ahead and merge. Thanks

Kraken/Test/asm/test_double_label.S
miriampolzer: Ah, good to know. This works on my machine, and the CI too.
Let me change it to two different labels.

Kraken/ParserTests.lean
miriampolzer: Yup, can be a followup. As mentioned in the other comment, my machine and CI accept this, your doesn't.

Kraken/Parser.lean
miriampolzer: yes, this is what made errors disappear.

Kraken/ParserTests.lean
miriampolzer: You're right, thanks for catching this.
The issue here is not the eof vs. newline though (bar eof works fine, as far as I have tested it).

The parser just doesn't support bare displacements yet. I had taken the parser at face value but ofc bare memory addresses should work too. Right now the parser reads 1 as displacement, and then looks for the opening bracket, doesn't find it, and fails.

Not that easy to test this though, because bare memory addresses won't work on real hw. But there's probably a way to hack this...
Anyways, adding this should be a separate change as it's non trivial. I will move the test down to the broken section and add a comment.

PR #117, opened 2026-05-21
Fix test with two labels

miriampolzer:
This should have been in #116, I forgot to add it. On Jonathan's machine, the test doesn't pass with two labels with the same name.


protz (approved):

protz:
Thank you!

PR #118, opened 2026-05-27
Complete proof for Executable.directivesFromStart; Step1 example

aferr:
Note: this change depends on my in-flight PR about parameterizing omnisemantics (#104 )

Finishes a proof for Executable.directivesFromStart.
This means we have truly complete examples rather than incomplete ones relying
on a sorried theorem.

Notably this involves refactoring Executable.withAddress into a definition
that supports proof by structural induction rather than the prior one based on
scanl.

This also adds an example for the step1 transition flavor of the omnisemantics.

The proof for `Executable.directivesAtStart` needed to complete the step1
example got a bit complicated so I moved it to Theorems.lean (and deleted
a definition that I left sorried).


aferr:
Ah... realize now @protz had this in-flight in a different way. This way does not require the version bump though does propose a semantics change to finish it...

Kraken/Semantics.lean
```diff
@@ -581,8 +581,12 @@ def scanl {α β} (f : β → α → β) (init : β) (as : List α) : List β :=
   Id.run <| Kraken.Compat.scanlM (pure <| f · ·) init as
 end Kraken.Compat
 
-def Executable.withAddresses (e : Executable)  : List (Int64 × Directive × Nat) :=
-  (Kraken.Compat.scanl (fun (p, _, _) (d, z) => (p+.ofNat z, d, z)) (e.1, .byteArray (.mk #[]), 0) e.2)
```
andres-erbsen, line 585: We discussed this change in the meeting and https://github.com/AeneasVerif/kraken/pull/102 seems preferable if feasible now

aferr: Converting this to draft since I am using this commit and just want this as essentially a bookmark if that's alright.

PR #119, opened 2026-06-05
[DONT REVIEW] Some Generally Useful Theorems

aferr:
These were made looking at a program involving some movs. Much of the bitvector stuff is specifically targeting the masks used in the semantics. The proofs have mostly been simplified to avoid too much brittleness.

There is also a file with some definitions I often use for specs about heap and stack region validity along with some theorems about heap validity. Likely many specs will want to make similar assumptions so the theorems around these may be of general use.

### DIFF
--- a/Kraken/SpecCommon.lean
+++ b/Kraken/SpecCommon.lean
@@ -1,0 +1,55 @@
+import Kraken.Semantics
+import Kraken.Theorems
+
+/-- Asserts that a contiguous region of n 64-bit words starting at addr exists in the heap. -/
+def heapRegionValid (s : MachineState) (addr : UInt64) (n : Nat) : Prop :=
+  ∀ i : Nat, i < n → (s.1.dmem[addr + i.toUInt64 * 8]?).isSome
+
+/-- Stack is valid: rsp is 8-byte aligned and has at least `minBytes` of space. -/
+def stackValid (s : MachineState) (minBytes : Nat) : Prop :=
+  s.1.regs.rsp.toNat % 8 = 0 ∧ s.1.regs.rsp.toNat ≥ minBytes
+
+
+-- ============================================================================
+-- heapRegionValid Theorems
+-- ============================================================================
+theorem heap_valid_base (s : MachineState) (addr : UInt64) (n : Nat)
+    (h_mem : heapRegionValid s addr n) (h_nz : 0 < n) :
+    (s.1.dmem[addr]?).isSome := by
+  have h_0 := h_mem 0 h_nz
+  have hy : addr + Nat.toUInt64 0 * 8 = addr := by
+    change addr + 0 = addr
+    simp
+  rw [hy] at h_0
+  exact h_0
+
+theorem dmem_get_of_heapRegionValid_base (s_data : MachineData) (pc : Int64) (base : UInt64) (N : Nat)
+    (h_mem : heapRegionValid (s_data, pc) base N) (h_bound : 0 < N := by decide) :
+    s_data.dmem[base]? = some ((s_data.dmem[base]?).get (by
+      have h_val := heap_valid_base (s_data, pc) base N h_mem h_bound
+      simp [h_val])) := by
+  have h_val := heap_valid_base (s_data, pc) base N h_mem h_bound
+  exact option_eq_some _ h_val
+
+theorem dmem_get_of_heapRegionValid (s_data : MachineData) (pc : Int64) (base : UInt64) (N : Nat)
+    (h_mem : heapRegionValid (s_data, pc) base N) (offset_bytes : UInt64) (offset_idx : Nat)
+    (h_eq : offset_bytes = offset_idx.toUInt64 * 8 := by decide)
+    (h_bound : offset_idx < N := by decide) :
+    s_data.dmem[base + offset_bytes]? = some ((s_data.dmem[base + offset_idx.toUInt64 * 8]?).get (by
+      have h_val := h_mem offset_idx h_bound
+      simp [h_val])) := by
+  have h_val := h_mem offset_idx h_bound
+  rw [h_eq]
+  exact option_eq_some _ h_val
+
+theorem heapRegionValid_get (s_data : MachineData) (pc : Int64) (base : UInt64) (N : Nat)
+    (h_mem : heapRegionValid (s_data, pc) base N) (offset : Nat) (h_bound : offset < N) :
+    (s_data.dmem[base + offset.toUInt64 * 8]?).isSome = true := by
+  have h := h_mem offset h_bound
+  exact h
+
+macro "resolve_mem_option" : tactic => `(tactic|
+  (
+    try rw [option_eq_some _ (heapRegionValid_get _ _ _ _ (by assumption) _ (by decide))]
+  )
+)
--- a/Kraken/Theorems.lean
+++ b/Kraken/Theorems.lean
@@ -310,14 +310,8 @@ theorem uint64_add_mask_align_8_val (x : UInt64) (offset : Nat)
     (UInt64.ofBitVec (BitVec.ofInt 64 (x.toBitVec.toInt + (offset : Int)))) &&& 18446744073709551608 =
     (UInt64.ofBitVec (BitVec.ofInt 64 (x.toBitVec.toInt + (offset : Int)))) := by
   apply uint64_ext
-  have h_align : BitVec.ofInt 64 (x.toBitVec.toInt + (offset : Int)) % 8#64 = 0#64 := by
-    apply uint64_add_align_bv_8 _ _ h_x h_off
-  have h_mask : BitVec.ofInt 64 (x.toBitVec.toInt + (offset : Int)) &&& 18446744073709551608#64 =
-      BitVec.ofInt 64 (x.toBitVec.toInt + (offset : Int)) := by
-    revert h_align
-    bv_decide
-  change (BitVec.ofInt 64 (x.toBitVec.toInt + (offset : Int)) &&& 18446744073709551608#64) = _
-  exact h_mask
+  apply bitvec_mask_align_8
+  apply uint64_add_align_bv_8 _ _ h_x h_off
 
 @[simp]
 theorem uint64_add_mask_align_8_val_int (x : UInt64) (offset : Int)
@@ -325,14 +319,8 @@ theorem uint64_add_mask_align_8_val_int (x : UInt64) (offset : Int)
     (UInt64.ofBitVec (BitVec.ofInt 64 (x.toBitVec.toInt + offset))) &&& 18446744073709551608 =
     (UInt64.ofBitVec (BitVec.ofInt 64 (x.toBitVec.toInt + offset))) := by
   apply uint64_ext
-  have h_align : BitVec.ofInt 64 (x.toBitVec.toInt + offset) % 8#64 = 0#64 := by
-    apply uint64_add_align_bv_8_int _ _ h_x h_off
-  have h_mask : BitVec.ofInt 64 (x.toBitVec.toInt + offset) &&& 18446744073709551608#64 =
-      BitVec.ofInt 64 (x.toBitVec.toInt + offset) := by
-    revert h_align
-    bv_decide
-  change (BitVec.ofInt 64 (x.toBitVec.toInt + offset) &&& 18446744073709551608#64) = _
-  exact h_mask
+  apply bitvec_mask_align_8
+  apply uint64_add_align_bv_8_int _ _ h_x h_off
 
 @[simp]
 theorem uint64_toNat_add_mask_align_8 (x : UInt64) (offset : Nat)
@@ -474,91 +462,18 @@ theorem mul_8_and_7 (k : Nat) : (8 * k) &&& 7 = 0 := by
 
 @[simp] theorem bitvec_add_mask_0 (rdi : UInt64) (h1 : rdi.toNat % 8 = 0) :
   rdi.toBitVec &&& ~~~7#64 = rdi.toBitVec := by
-  apply BitVec.eq_of_toNat_eq
-  have h3 : (~~~7#64).toNat = 18446744073709551608 := rfl
-  have h4 : (rdi.toBitVec &&& ~~~7#64).toNat = rdi.toBitVec.toNat &&& 18446744073709551608 := by rfl
-  rw [h4]
-  have hk1 : rdi.toBitVec.toNat = rdi.toNat := rfl
-  have hk2 : rdi.toNat = 8 * (rdi.toNat / 8) := by omega
-  have hv : rdi.toBitVec.toNat % 8 = 0 := by
-    rw [hk1, hk2]
-    omega
-  have hk : rdi.toBitVec.toNat = 8 * (rdi.toBitVec.toNat / 8) := by omega
-  have hv_lt : 8 * (rdi.toBitVec.toNat / 8) < 18446744073709551616 := by rw [← hk]; exact rdi.toBitVec.isLt
-  generalize (rdi.toBitVec.toNat / 8) = k at hk hv_lt
-  rw [hk]
-  apply Nat.eq_of_testBit_eq
-  intro i
-  simp
-  cases i with
-  | zero =>
-    have : (8 * k).testBit 0 = false := by simp [Nat.testBit]; omega
-    simp [this]
-  | succ i => cases i with
-    | zero =>
-      have : (8 * k).testBit 1 = false := by simp [Nat.testBit]; omega
-      simp [this]
-    | succ i => cases i with
-      | zero =>
-        have : (8 * k).testBit 2 = false := by simp [Nat.testBit]; omega
-        simp [this]
-      | succ i =>
-        by_cases hi : i < 61
-        · have h_tb : Nat.testBit 18446744073709551608 (i + 3) = true := by
-            have : ∀ j : Fin 61, Nat.testBit 18446744073709551608 (j.val + 3) = true := by decide
-            exact this ⟨i, hi⟩
-          simp [h_tb]
-        · have h_tb : (8 * k).testBit (i + 3) = false := by
-            apply Nat.testBit_lt_two_pow
-            have : 8 * k < 2^64 := hv_lt
-            have : 2^64 ≤ 2^(i + 3) := Nat.pow_le_pow_right (by decide) (by omega)
-            omega
-          simp [h_tb]
+  have : rdi.toBitVec % 8#64 = 0#64 := uint64_align_bv_8 _ h1
+  revert this
+  bv_decide
 
 @[simp] theorem bitvec_add_mask_c (rdi : UInt64) (c : BitVec 64) (h1 : rdi.toNat % 8 = 0) (h2 : c.toNat % 8 = 0 := by decide) :
   (rdi.toBitVec + c) &&& ~~~7#64 = rdi.toBitVec + c := by
-  apply BitVec.eq_of_toNat_eq
-  have h3 : (~~~7#64).toNat = 18446744073709551608 := rfl
-  have h4 : ((rdi.toBitVec + c) &&& ~~~7#64).toNat = (rdi.toBitVec + c).toNat &&& 18446744073709551608 := by rfl
-  rw [h4]
-  have hk1 : rdi.toBitVec.toNat = rdi.toNat := rfl
-  have hk2 : rdi.toNat = 8 * (rdi.toNat / 8) := by omega
-  have hk3 : c.toNat = 8 * (c.toNat / 8) := by omega
-  have hv : (rdi.toBitVec + c).toNat % 8 = 0 := by
-    change (rdi.toBitVec.toNat + c.toNat) % 18446744073709551616 % 8 = 0
-    rw [hk1, hk2, hk3]
-    omega
-  have hk : (rdi.toBitVec + c).toNat = 8 * ((rdi.toBitVec + c).toNat / 8) := by omega
-  have hv_lt : 8 * ((rdi.toBitVec + c).toNat / 8) < 18446744073709551616 := by rw [← hk]; exact (rdi.toBitVec + c).isLt
-  generalize ((rdi.toBitVec + c).toNat / 8) = k at hk hv_lt
-  rw [hk]
-  apply Nat.eq_of_testBit_eq
-  intro i
-  simp
-  cases i with
-  | zero =>
-    have : (8 * k).testBit 0 = false := by simp [Nat.testBit]; omega
-    simp [this]
-  | succ i => cases i with
-    | zero =>
-      have : (8 * k).testBit 1 = false := by simp [Nat.testBit]; omega
-      simp [this]
-    | succ i => cases i with
-      | zero =>
-        have : (8 * k).testBit 2 = false := by simp [Nat.testBit]; omega
-        simp [this]
-      | succ i =>
-        by_cases hi : i < 61
-        · have h_tb : Nat.testBit 18446744073709551608 (i + 3) = true := by
-            have : ∀ j : Fin 61, Nat.testBit 18446744073709551608 (j.val + 3) = true := by decide
-            exact this ⟨i, hi⟩
-          simp [h_tb]
-        · have h_tb : (8 * k).testBit (i + 3) = false := by
-            apply Nat.testBit_lt_two_pow
-            have : 8 * k < 2^64 := hv_lt
-            have : 2^64 ≤ 2^(i + 3) := Nat.pow_le_pow_right (by decide) (by omega)
-            omega
-          simp [h_tb]
+  have hv : (rdi.toBitVec + c) % 8#64 = 0#64 := by
+    apply bitvec_add_align_8
+    · exact uint64_align_bv_8 _ h1
+    · exact BitVec.eq_of_toNat_eq (show c.toNat % 8 = 0 from h2)
+  revert hv
+  bv_decide
 
 -- ============================================================================
 -- UInt64 Address Distinctness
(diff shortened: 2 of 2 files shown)


Kraken/SpecCommon.lean
```diff
@@ -0,0 +1,55 @@
+import Kraken.Semantics
+import Kraken.Theorems
+
+/-- Asserts that a contiguous region of n 64-bit words starting at addr exists in the heap. -/
+def heapRegionValid (s : MachineState) (addr : UInt64) (n : Nat) : Prop :=
```
aferr, line 5: BTW it's totally fine if you don't want this file upstreamed. Just thought I'd offer.

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+-- ============================================================================
+-- BitVec Conversions & Simplifications
+-- ============================================================================
+
+theorem ofInt_add (z1 z2 : Int) :
+    BitVec.ofInt 64 (z1 + z2) = BitVec.ofInt 64 z1 + BitVec.ofInt 64 z2 := by
+  apply BitVec.eq_of_toNat_eq
+  dsimp only [BitVec.ofInt, HAdd.hAdd, Add.add, BitVec.add, BitVec.toNat]
+  simp
+  omega
```
andres-erbsen, line 32: https://leanprover-community.github.io/mathlib4_docs/Init/Data/BitVec/Lemmas.html#BitVec.ofInt_add

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+
+theorem ofInt_toInt (x : BitVec 64) :
+    BitVec.ofInt 64 x.toInt = x := by
+  apply BitVec.eq_of_toNat_eq
+  dsimp only [BitVec.ofInt, BitVec.toInt, BitVec.toNat]
+  simp
+  omega
```
andres-erbsen, line 39: https://leanprover-community.github.io/mathlib4_docs/Init/Data/BitVec/Lemmas.html#BitVec.ofInt_toInt

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+  | none => contradiction
+  | some v => rfl
+
+@[simp] theorem isSome_to_exists {α} (o : Option α) (h : o.isSome) : ∃ v, o = some v := by
+  cases o
+  · contradiction
+  · exact ⟨_, rfl⟩
+
+-- ============================================================================
+-- BitVec Conversions & Simplifications
```
andres-erbsen, line 24: `namespace BitVec`

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
 -/
 
 import Kraken.Semantics
+import Kraken.Tactics
+
+-- ============================================================================
+-- Option Lemmas
+-- ============================================================================
+
+theorem option_eq_some {α} (opt : Option α) (h : opt.isSome = true) :
```
andres-erbsen, line 12: https://leanprover-community.github.io/mathlib4_docs/Init/Data/Option/Basic.html#Option.some_get

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+    opt = some (opt.get (by simp [h])) := by
+  cases opt with
+  | none => contradiction
+  | some v => rfl
+
+@[simp] theorem isSome_to_exists {α} (o : Option α) (h : o.isSome) : ∃ v, o = some v := by
```
andres-erbsen, line 18: duplicate of previous lemma, or BitVec.add_right_eq_self if you insist on having an existential

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+  omega
+
+@[simp] theorem ofInt_signed (v : BitVec 64) : BitVec.ofInt 64 v.signed = v := by
+  apply BitVec.eq_of_toNat_eq
+  simp [BitVec.signed, BitVec.toInt, BitVec.ofInt]
+  omega
+
+@[simp]
+theorem ofInt_toInt_add (x : BitVec 64) (offset : Nat) :
+    BitVec.ofInt 64 (x.toInt + (offset : Int)) = x + BitVec.ofNat 64 offset := by
```
andres-erbsen, line 48: as the proof shows, this lemma is a trivial combination of two other simp-friendly lemmas

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+  rw [ofInt_add, ofInt_toInt]
+  rfl
+
+@[simp]
+theorem ofInt_toInt_add_int (x : BitVec 64) (offset : Int) :
+    BitVec.ofInt 64 (x.toInt + offset) = x + BitVec.ofInt 64 offset := by
+  rw [ofInt_add, ofInt_toInt]
```
andres-erbsen, line 55: ditto

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+
+theorem toNat_ofInt_eq (z : Int) :
+    ((z % 18446744073709551616).toNat) = (BitVec.ofInt 64 z).toNat := by
+  rfl
```
andres-erbsen, line 59: BitVec.toNat_ofInt

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+
+theorem ofInt_mod_8 (offset : Int) (h : offset % 8 = 0) :
+    (BitVec.ofInt 64 offset).toNat % 8 = 0 := by
```
andres-erbsen, line 62: Nat.mod_mod_of_dvd and previously mentioned injection lemmas

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+  dsimp [BitVec.ofInt]
+  have h_ge : offset % 18446744073709551616 ≥ 0 := by
+    apply Int.emod_nonneg
+    decide
+  omega
+
+theorem ofInt_inj_int (o1 o2 : Int) (h_o1 : 0 ≤ o1 ∧ o1 < 18446744073709551616)
+    (h_o2 : 0 ≤ o2 ∧ o2 < 18446744073709551616) (h : BitVec.ofInt 64 o1 = BitVec.ofInt 64 o2) :
+    o1 = o2 := by
```
andres-erbsen, line 71: `BitVec.toInt_ofInt` and grind

(This is a special case of what an intblaster would do. We should seek to adopt an intblaster.)

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+  have h_o1_eq : (BitVec.ofInt 64 o1).toNat = (o1 % 18446744073709551616).toNat := rfl
+  have h_o2_eq : (BitVec.ofInt 64 o2).toNat = (o2 % 18446744073709551616).toNat := rfl
+  rw [h_o1_eq, h_o2_eq] at h_eq
+  have h1 : o1 % 18446744073709551616 = o1 := by omega
+  have h2 : o2 % 18446744073709551616 = o2 := by omega
+  rw [h1, h2] at h_eq
+  omega
+
+
+@[simp] theorem bitvec_extractLsb'_0_64 (x : BitVec 64) : BitVec.extractLsb' 0 64 x = x := by bv_decide
```
andres-erbsen, line 82: [BitVec.extractLsb'_eq_self](https://leanprover-community.github.io/mathlib4_docs/Init/Data/BitVec/Lemmas.html#BitVec.extractLsb'_eq_self)

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+
+
+@[simp] theorem bitvec_zeroExtend_64_64 (x : BitVec 64) : BitVec.zeroExtend 64 x = x := by bv_decide
```
andres-erbsen, line 85: [BitVec.shiftLeftZeroExtend_eq](https://leanprover-community.github.io/mathlib4_docs/Init/Data/BitVec/Lemmas.html#BitVec.shiftLeftZeroExtend_eq), 
BitVec.setWidth_setWidth_eq_self

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+
+
+@[simp]
```
andres-erbsen, line 88: Let's not add anything to the global simpset

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+theorem bitvec_replace_all_64 (old : BitVec 64) (new : BitVec 64) :
+    BitVec.replace old 0 new = new := by
+    dsimp only [BitVec.replace, BitVec.drop, BitVec.take]
+    bv_decide
```
andres-erbsen, line 92:
```suggestion
    theorem replace_all {w} (old : BitVec w) (new : BitVec w) :
        BitVec.replace old 0 new = new := by
      dsimp only [BitVec.replace, BitVec.drop, BitVec.take]
      rw [Nat.zero_add]
      rw [Nat.sub_self]
      simp
```

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+
+
+theorem bitvec_add_left_cancel (x y z : BitVec 64) (h : x + y = x + z) : y = z := by
```
andres-erbsen, line 95: if you have access to mathlib, this is just `add_left_cancel`. and without it mathlib it's by grind or ring, no bitblasting needed. If tou want to add it still, rename to match Stdlib conventions.

Kraken/Theorems.lean
```diff
@@ -3,9 +3,104 @@ Kraken - Helper Theorems
...
+  revert h
+  bv_decide
+
+-- ============================================================================
+-- UInt64 Arithmetic Fundamentals
```
andres-erbsen, line 100: Please follow Stdlib conventions wrt namespaces because these affect all users of this file.

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+  have h3 : some (getElem (m.insert k v) k' h1) = (m.insert k v).get? k' := by simp
+  have h4 : some (getElem m k' h2) = m.get? k' := by simp
+  rw [h3, h4]
+  simp [h_ne]
+
+-- ============================================================================
+-- MachineState Projections
+-- ============================================================================
+
+@[simp] theorem md_regs_proj (r s d) : ({ regs := r, status := s, dmem := d } : MachineData).regs = r := rfl
```
andres-erbsen, line 633: We should have a simp config or simproc do this generically. @protz do you have something like this alreayd?

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+
+@[simp] theorem setReg_regs (s : MachineData) {w} (r : Reg w) (v : w.type) :
+  (s.setReg r v).regs = s.regs.set r v := rfl
+
+@[simp] theorem setReg_dmem (s : MachineData) {w} (r : Reg w) (v : w.type) :
+  (s.setReg r v).dmem = s.dmem := rfl
+
+@[simp] theorem setReg_status (s : MachineData) {w} (r : Reg w) (v : w.type) :
+  (s.setReg r v).status = s.status := rfl
```
andres-erbsen, line 642: Where are these useful? a little bit of unfolding would also do in any use case I can think of

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+@[simp] theorem ext_hash_map_getElem_insert_ne (m : Std.ExtHashMap UInt64 UInt64) (k k' : UInt64) (v : UInt64) (h_ne : k ≠ k') (h1) (h2) :
+  getElem (m.insert k v) k' h1 = getElem m k' h2 := by
+  apply Option.some.inj
+  have h3 : some (getElem (m.insert k v) k' h1) = (m.insert k v).get? k' := by simp
+  have h4 : some (getElem m k' h2) = m.get? k' := by simp
+  rw [h3, h4]
+  simp [h_ne]
+
+-- ============================================================================
+-- MachineState Projections
```
andres-erbsen, line 630: If any lemmas remain here, put them in the right namespace

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+  · intro h h_eq; apply h; apply UInt64.toNat_inj.1; have h2 := congrArg UInt64.toNat h_eq; revert h2; change (a.toNat + b.toNat) % 18446744073709551616 = a.toNat → (b.toNat = 0); intro h; have h3 := UInt64.toNat_lt_size a; have h4 := UInt64.toNat_lt_size b; change a.toNat < 18446744073709551616 at h3; change b.toNat < 18446744073709551616 at h4; omega
+
+@[simp] theorem add_inj_right_ne_self (a b : UInt64) : (a ≠ a + b) ↔ (b ≠ 0) := by
+  constructor
+  · intro h h_eq; apply h; apply UInt64.toNat_inj.1; have h2 := congrArg UInt64.toNat h_eq; revert h2; change b.toNat = 0 → a.toNat = (a.toNat + b.toNat) % 18446744073709551616; intro h; have h3 := UInt64.toNat_lt_size a; have h4 := UInt64.toNat_lt_size b; change a.toNat < 18446744073709551616 at h3; change b.toNat < 18446744073709551616 at h4; omega
+  · intro h h_eq; apply h; apply UInt64.toNat_inj.1; have h2 := congrArg UInt64.toNat h_eq; revert h2; change a.toNat = (a.toNat + b.toNat) % 18446744073709551616 → (b.toNat = 0); intro h; have h3 := UInt64.toNat_lt_size a; have h4 := UInt64.toNat_lt_size b; change a.toNat < 18446744073709551616 at h3; change b.toNat < 18446744073709551616 at h4; omega
```
andres-erbsen, line 603: `≠` is just a notation for not of `=`, so these lemmas are congruence-wrapped duplicates of previous ones

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+    (h_o1 : off1 < 18446744073709551616 := by decide)
+    (h_o2 : off2 < 18446744073709551616 := by decide)
+    (h_ne : off1 ≠ off2 := by decide) :
+    base + UInt64.ofNat off1 ≠ base + UInt64.ofNat off2 := by
+  intro h
+  have h_false : (base + UInt64.ofNat off1 = base + UInt64.ofNat off2) = False := base_offset_eq_false base off1 off2 h_o1 h_o2 h_ne
+  rw [h] at h_false
+  simp at h_false
+
+@[simp] theorem add_inj_left_eq_self (a b : UInt64) : (a + b = a) ↔ (b = 0) := by
```
andres-erbsen, line 580: BitVec.add_right_eq_self

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+@[simp] theorem ext_hash_map_getElem_insert_eq (m : Std.ExtHashMap UInt64 UInt64) (k : UInt64) (v : UInt64) (h) :
+  getElem (m.insert k v) k h = v := by
+  simp
+@[simp] theorem ext_hash_map_getElem_insert_ne (m : Std.ExtHashMap UInt64 UInt64) (k k' : UInt64) (v : UInt64) (h_ne : k ≠ k') (h1) (h2) :
+  getElem (m.insert k v) k' h1 = getElem m k' h2 := by
+  apply Option.some.inj
+  have h3 : some (getElem (m.insert k v) k' h1) = (m.insert k v).get? k' := by simp
+  have h4 : some (getElem m k' h2) = m.get? k' := by simp
+  rw [h3, h4]
+  simp [h_ne]
```
andres-erbsen, line 627: Are these lemmas somehow better than the stdlib equivalents used to prove them?

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+  · intro h; apply UInt64.toNat_inj.1; have h2 := congrArg UInt64.toNat h; revert h2; change b.toNat = 0 → (a.toNat + b.toNat) % 18446744073709551616 = a.toNat; intro h; have h3 := UInt64.toNat_lt_size a; have h4 := UInt64.toNat_lt_size b; change a.toNat < 18446744073709551616 at h3; change b.toNat < 18446744073709551616 at h4; omega
+
+@[simp] theorem add_inj_right_eq_self (a b : UInt64) : (a = a + b) ↔ (b = 0) := by
+  constructor
+  · intro h; apply UInt64.toNat_inj.1; have h2 := congrArg UInt64.toNat h; revert h2; change a.toNat = (a.toNat + b.toNat) % 18446744073709551616 → (b.toNat = 0); intro h; have h3 := UInt64.toNat_lt_size a; have h4 := UInt64.toNat_lt_size b; change a.toNat < 18446744073709551616 at h3; change b.toNat < 18446744073709551616 at h4; omega
+  · intro h; apply UInt64.toNat_inj.1; have h2 := congrArg UInt64.toNat h; revert h2; change b.toNat = 0 → a.toNat = (a.toNat + b.toNat) % 18446744073709551616; intro h; have h3 := UInt64.toNat_lt_size a; have h4 := UInt64.toNat_lt_size b; change a.toNat < 18446744073709551616 at h3; change b.toNat < 18446744073709551616 at h4; omega
+
+@[simp] theorem add_inj_left_ne (a b c : UInt64) : (a + b ≠ a + c) ↔ (b ≠ c) := by
```
andres-erbsen, line 590: `grind` can prove this, are you sure we want the lemma?

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
   have hzero : (0 : Nat) % 2^64 = 0 := Nat.zero_mod (2^64)
   rw [h2, hkmod, hzero] at h1
   exact h_ne h1
+
+theorem uint64_ext (a b : UInt64) (h : a.toBitVec = b.toBitVec) : a = b := by
```
andres-erbsen, line 122: UInt64.eq_of_toBitVec_eq

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+  congr
+
+theorem ofNat_inj (o1 o2 : Nat) (h_o1 : o1 < 18446744073709551616) (h_o2 : o2 < 18446744073709551616)
+    (h : BitVec.ofNat 64 o1 = BitVec.ofNat 64 o2) : o1 = o2 := by
+  have h_eq : (BitVec.ofNat 64 o1).toNat = (BitVec.ofNat 64 o2).toNat := by rw [h]
+  simp only [BitVec.toNat_ofNat] at h_eq
+  omega
+
+@[simp] theorem uint64_mk_bv (v : BitVec 64) :
+  ({ toBitVec := v } : UInt64) = UInt64.ofBitVec v := by
```
andres-erbsen, line 134: We should go out of our way to avoid exposing this construction in proofs

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+
+theorem uint64_ext (a b : UInt64) (h : a.toBitVec = b.toBitVec) : a = b := by
+  cases a; cases b
+  dsimp [UInt64.toBitVec] at h
+  congr
+
+theorem ofNat_inj (o1 o2 : Nat) (h_o1 : o1 < 18446744073709551616) (h_o2 : o2 < 18446744073709551616)
+    (h : BitVec.ofNat 64 o1 = BitVec.ofNat 64 o2) : o1 = o2 := by
+  have h_eq : (BitVec.ofNat 64 o1).toNat = (BitVec.ofNat 64 o2).toNat := by rw [h]
+  simp only [BitVec.toNat_ofNat] at h_eq
```
andres-erbsen, line 130: this stdlib lemma is what i recommend using instead

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+  omega
+
+@[simp] theorem uint64_mk_bv (v : BitVec 64) :
+  ({ toBitVec := v } : UInt64) = UInt64.ofBitVec v := by
+  rfl
+
+@[simp] theorem uint64_add_c_general (rdi : UInt64) (c : Int) :
+  ({ toBitVec := BitVec.ofInt 64 (rdi.toBitVec.signed + c) } : UInt64) = rdi + (UInt64.ofBitVec (BitVec.ofInt 64 c)) := by
```
andres-erbsen, line 138: ditto

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+  apply BitVec.eq_of_toNat_eq
+  simp [BitVec.signed, BitVec.toInt, BitVec.ofInt]
+  omega
+
+-- ============================================================================
+-- Int64 / Int Conversion Lemmas
+-- ============================================================================
+
+@[simp] theorem int64_toint_c_add_general (rdi : UInt64) (c : Int) :
+  BitVec.zeroExtend 64 (BitVec.ofInt 64 ((BitVec.extractLsb' 0 64 rdi.toBitVec).toInt + 0 + c)) = rdi.toBitVec + BitVec.ofInt 64 c := by
```
andres-erbsen, line 151: the LHS itself could be simplified using simpler lemmas, let's not duplicate that in this lemma

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+  simp [BitVec.toInt, BitVec.ofInt]
+  have h_toInt : rdi.toBitVec.toInt = if 2 * rdi.toBitVec.toNat < 18446744073709551616 then (rdi.toBitVec.toNat : Int) else (rdi.toBitVec.toNat : Int) - 18446744073709551616 := rfl
+  omega
+
+-- Bridge lemma: lets simp prove (n : UInt64).toNat % 8 = 0 for concrete n.
+-- simp chains: uint64_mask_align_8 needs alignment → uint64_add_align_8 propagates it
+-- → this lemma reduces UInt64 literal alignment to Nat alignment (n % 8 = 0)
+-- → Nat.reduceMod computes n % 8 for ground n.
+@[simp] theorem uint64_ofNat_toNat_mod_8 (n : Nat) (h : n % 8 = 0) :
+    (OfNat.ofNat n : UInt64).toNat % 8 = 0 := by
```
andres-erbsen, line 166: again, this is modulo after modulo with a divisor like above

Kraken/Theorems.lean
```diff
@@ -23,3 +118,540 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
+  change n % 18446744073709551616 % 8 = 0
+  omega
+
+-- ============================================================================
+-- 8-Byte Alignment & Bitmask Lemmas
+-- ============================================================================
+
+@[simp]
+theorem uint64_align_bv_8 (x : UInt64) (h : x.toNat % 8 = 0) : x.toBitVec % 8#64 = 0#64 := by
```
andres-erbsen, line 175: We should go out of our way to keep all alignment definitions on the same type, probably Nat. Where does this come up?

andres-erbsen (changes requested):
Most lemmas added in this PR fall into two categories:

1. Memory-address manipulation. These are tedious, and I can see why you'd want to cache them, but  I think we should pivot to doing separation-logic right away instead of trying to hack around it. It really would save time, even in the short term. I did not comment on the parts that I believe would be superseded by that. Please take a look at https://github.com/AeneasVerif/kraken/pull/120 and see whether it looks helpful / what else you'd need immediately.
2. Duplication for the standard library. Let's teach our vibe-coding tools to avoid this in the future -- a one-liner for how to use the Loogle API seems to help already.

So my sense is that we probably don't want to move forward with this, but I'm open to being persuaded.

aferr:
> 1. Memory-address manipulation. These are tedious, and I can see why you'd want to cache them, but  I think we should pivot to doing separation-logic right away instead of trying to hack around it. It really would save time, even in the short term. I did not comment on the parts that I believe would be superseded by that. Please take a look at #120 and see whether it looks helpful / what else you'd need immediately.


A common category of these lemmas is to prove address alignment. Most of the ones involving constants are about the specific masks used in the semantics. I don't believe a standard implementation of separation logic would address alignment or the bit vector operations involved to prove it. Were you planning to develop some machinery to prove address alignment as a part of your separation logic work?

aferr:
> > 1. Memory-address manipulation. These are tedious, and I can see why you'd want to cache them, but  I think we should pivot to doing separation-logic right away instead of trying to hack around it. It really would save time, even in the short term. I did not comment on the parts that I believe would be superseded by that. Please take a look at [start separation logic #120](https://github.com/AeneasVerif/kraken/pull/120) and see whether it looks helpful / what else you'd need immediately.
> 
> A common category of these lemmas is to prove address alignment. Most of the ones involving constants are about the specific masks used in the semantics. I don't believe a standard implementation of separation logic would address alignment or the bit vector operations involved to prove it. Were you planning to develop some machinery to prove address alignment as a part of your separation logic work?

At a very quick glance, what you're doing specifically might move our joined project forward more than my image of separation logic in general. Will take a deeper look shortly.

Kraken/Theorems.lean
protz: no but I think we need a kraken simpset

aferr:
We do no intend to merge this. I'd like to keep this open just to use as a bookmark for myself for a bit longer if that's ok

Issue #41, opened 2026-03-27
[tracking issue] Separation logic for memory


andres-erbsen:
It's come up a few times that we intend to port a memory-only, first-order, automation-oriented separation logic from past work in Rocq. The smallest example that needs this is commuting read and write accesses through different pointers during verification. We'll also want to use the same framework to calculate frame (unchanged-memory) predicates on function calls and other control-flow merge points.

[Here](https://gist.github.com/andres-erbsen/08ae50e37c04f281c619497f91db770f) is a sketch of what the specifications might look like based on Separation.v in coqutil. Feedback welcome.

protz:
@samuelgruetter how does your proposal relate to what's currently in the repository after landing Andres' initial proposal?

PR #120, opened 2026-06-11
start separation logic

andres-erbsen:
To be squashed.

Layers of the representations, for the example of UInt64:
- `bs := UInt64.toBytes x : list byte`
- `m := List.At bs a : ExtHashMap UIn64 byte`
- `Eq m : ExtHashMap UIn64 byte -> Prop`
- `sep : (mem -> Prop) -> (mem -> Prop) -> (mem -> Prop)`

It's generally advisable to reason on the *input* side of these adapters, without even mentioning the later ones if their properties are not relevant. For example, concatenating representations of two UInt16s would give a representation of an UInt32, and the value represented can be derived using the byte-list level only, without considering memory or separation.

- https://github.com/AeneasVerif/kraken/issues/41
- https://raw.githubusercontent.com/mit-plv/coqutil/master/src/coqutil/Map/SeparationLogic.v
- https://raw.githubusercontent.com/mit-plv/coqutil/master/src/coqutil/Map/SeparationMemory.v


aferr:
Thanks for your detailed feedback on the memory lemmas I was working on, your attention to my use case and putting this together!

I think this could be the way forward, but I do have serious reservations about proof complexity.  This change removes the need to reason about bit masks, but in doing so removes the constraint that memory access is aligned and necessitates separation logic to reason about overlap (or not) of memory. So this has tradeoffs rather than being pure proof upside. (Towards our joint project: there are  there are no applications where we strictly need unaligned memory, so it's just a question of reducing proof difficulty). 

To help us ensure we have the machinery we need to manage the greater generality the semantics here enables, can you completely work proofs for a few small and deliberately chosen example programs? These are the ones I have in mind.

1. A basic example moving registers in and out of memory:
```
    movq %rax, (%rdi)
    movq %rcx, 8(%rdi)
    movq (%rdi), %r12
    movq 8(%rdi), %r13
```

proving
```
    Eventually (straightlineStep (layout move_2_regs_to_heap))
      (fun s' =>
        s'.1.regs.r12 = s₀.regs.rax ∧
        s'.1.regs.r13 = s₀.regs.rcx ∧
        s'.1.regs.rdi = s₀.regs.rdi)
      (s₀, layout.start) := by
```

2. An example using Scale-Index Base Addressing as in:
```
    movq $42, %rax
    movq %rax, (%rdi, %r15, 8)
    movq $0, %rax
    movq (%rdi, %r15, 8), %rax
```

just proving that in the final state rax is 42 (with your Eventually construct)

3. An example involving dynamic stack allocation:
```
    movq $99, -8(%rsp)
    movq %rsp, %rbp
    leaq -1024(%rsp, %r9, 8), %rsp
    movq $42, %rax
    movq %rax, 16(%rsp, %r15, 8)
    movq $0, %rax
    movq 16(%rsp, %r15, 8), %rax
    movq %rbp, %rsp
    movq -8(%rsp), %rbx
 ```
 
 proving: `s'.1.regs.rax = 42 ∧ s'.1.regs.rbx = 99 ∧ s'.1.regs.rsp = s₀.regs.rsp`
 (where s' is final state, s0 is initial)
 
 4. An example that mimics operating on a large in-memory data structure in ALU ops by just having a big offset and doing some arithmetic like the following:
 ```
     movq $42, %rax
    movq %rax, 136(%rdx)
    movq $100, %rcx
    addq 136(%rdx), %rcx
 ```
 
Without some worked out examples like these, I am concerned about whether we are progressing or just making the challenges different. However, given that Kraken's examples don't have many programs involving memory, if we can both build out this separation logic solution and work out these examples (chosen based on our joint project) I think this would be clear progress for both Kraken's overall infrastructure and our joint project.

**Comment by andres-erbsen** [comment]
PTAL, there are some rather sloppy attemtps at what you asked for in Examples.lean. Let's discuss at the meeting what to do next here and how to distribute the work between people.

File: Kraken/ToBytes.lean [hunk 1]
```diff
@@ -0,0 +1,533 @@
+abbrev Nat.take (v : Nat) (w : Nat) : Nat := v % 2 ^ w
+abbrev Int.take (v : Int) (w : Nat) : Int := v % (2 : Int) ^ w
+@[simp] theorem pow_two_eight : (2 : Int) ^ (8 : Nat) = 256 := by rfl
+@[simp] theorem Nat.pow_two_eight : (2 : Nat) ^ (8 : Nat) = 256 := by rfl
```
andres-erbsen: TODO: remove this

Kraken/Theorems.lean
```diff
@@ -25,50 +30,9 @@ theorem UInt64_ofInt_natCast_ne_zero (k : Nat) (h_lt : k < 2^64) (h_ne : k ≠ 0
...
-    bv_decide
-  rw [this]
-  have: addr &&& 7#64 = 0 := by bv_decide
-  simp [this,Width.bits]
-  rw [Std.ExtHashMap.getElem?_eq_some_getElem! hContains]
+theorem BitVec.take_all {w : Nat} (x : BitVec w) : x.take w = x := by
+  simp [BitVec.take]
+
+theorem BitVec.take_self_64 (x : BitVec 64) : x.take 64 = x :=
+  BitVec.take_all x
```
andres-erbsen: TODO: remove

Kraken/Syntax.lean
```diff
@@ -7,8 +7,8 @@ instance : ToString Width where
   toString | .W8 => "w8" | .W16 => "w16" | .W32 => "w32" | .W64 => "w64"
 
 namespace Width
-def bits : Width → Nat | W8 => 8 | W16 => 16 | W32 => 32 | W64 => 64
-def bytes : Width → Nat | W8 => 1 | W16 => 2 | W32 => 4 | W64 => 8
+@[simp, reducible] def bits : Width → Nat | W8 => 8 | W16 => 16 | W32 => 32 | W64 => 64
+@[simp, reducible] def bytes : Width → Nat | W8 => 1 | W16 => 2 | W32 => 4 | W64 => 8
```
andres-erbsen, line 11: Any objections to this?

Kraken/SeparationPredicates.lean
```diff
@@ -0,0 +1,26 @@
...
+
+def ptsto (k : key) (v : value) : map key value → Prop := Eq (.insert ∅ k v)
+
+def flex (p : map key value → Prop) (m : map key value) : Prop := ∃ q, sep p q m
+
+infixl:60 " ⋆ " => sep
+notation:70 m " =⋆ " P => ((P : ExtHashMap _ _ → Prop) m : Prop)
+notation:70 m " ≥⋆ " P => flex P m
+infixl:35 " ∧ " => and
+infixl:35 " ∧≥ " => fun a b => and a (flex b)
```
andres-erbsen: Can anyone think of a good way to validate the level choices?

Kraken/Mem.lean
```diff
@@ -0,0 +1,212 @@
...
+  (loadBytes m a n).map Int.ofBytes
+
+
+def Mem.storeBytes {w : Nat} (m : Mem w) (a : BitVec w) (bs : List UInt8) : Mem w :=
+  m.union (bs.At a)
+
+def Mem.storeInt {w : Nat} (m : Mem w) (a : BitVec w) (n : Nat) (v : Int) : Mem w :=
+  storeBytes m a (Int.toBytes n v)
+
+private theorem List.zipIdx_pairwise_lt {α} (l : List α) (k : Nat) :
```
andres-erbsen: It would be nice to have a low-thinking process for lemmas that "ought to be in Sdlib": gathering, triage, upstreaming, removing.

Kraken/Examples/Examples.lean
```diff
@@ -430,3 +407,223 @@ example [layout : Layout] (s: MachineData)
...
+  rw [h_addr_eq] at h_mem
+  rw [sep_comm (Eq ((stack.take 1016).At (rsp.toBitVec - 1024#64)))] at h_mem
+  rw [sep_assoc] at h_mem
+  have h_mem1 := Mem.storeInt_sep (rsp.toBitVec + 18446744073709551608#64) 8 (stack.drop 1016) (R ⋆ Eq ((stack.take 1016).At (rsp.toBitVec - 1024#64))) mem ⟨h_mem, h_len_drop⟩ 99
+  change ss.dmem =⋆ (stack.drop 1016).At (rsp.toBitVec + 18446744073709551608#64) ⋆ (R ⋆ Eq ((stack.take 1016).At (rsp.toBitVec - 1024#64))) at h_mem
+  rw [store_sep ss _ .W64 _ _ (stack.drop 1016) (R ⋆ Eq ((stack.take 1016).At (rsp.toBitVec - 1024#64)))]
+  case h_mem => exact h_mem
+  case h_len => exact h_len_drop
+  sym =>
+  -- FIXME: kstep here takes too long
```
andres-erbsen, line 595: I did not troubleshoot this too deeply, just where I'm ending my day

Kraken/ToBytes.lean
```diff
@@ -0,0 +1,533 @@
...
+    rw [h_pow]
+    have h_b_lt : (b.toNat : Int) < 256 := by
+      cases b with | ofBitVec val =>
+      simp [UInt8.toNat]
+      have := val.isLt
+      omega
+    have h_ge := Int.ofBytes_ge_zero bs
+    omega
+
+private theorem emod_div_mul (x : Int) (M : Int) (d : Int) (hM : M > 0) (hd : d > 0) :
```
andres-erbsen, line 130: i am guessing applying a human brain here would find a way to use stdlib lemmas instead of this one

Kraken/Separation/Lemmas.lean
```diff
@@ -0,0 +1,301 @@
+import Kraken.SeparationPredicates
+
+open Std
+open Std.ExtHashMap
+
+namespace Std.ExtHashMap
+
+variable {key value : Type} [BEq key] [EquivBEq key] [Hashable key] [LawfulHashable key] [LawfulBEq key]
+
+omit [LawfulBEq key] in
```
andres-erbsen: I don't know whether its right to omit the instances. Does Stdlib have a policy for how they choose which instances to require?

Kraken/SeparationPredicates.lean
```diff
@@ -0,0 +1,26 @@
...
+
+def and (p q : map key value → Prop) (m : map key value) : Prop :=
+  p m ∧ q m
+
+def ptsto (k : key) (v : value) : map key value → Prop := Eq (.insert ∅ k v)
+
+def flex (p : map key value → Prop) (m : map key value) : Prop := ∃ q, sep p q m
+
+infixl:60 " ⋆ " => sep
+notation:70 m " =⋆ " P => ((P : ExtHashMap _ _ → Prop) m : Prop)
```
aferr: Is this magic wand? If so maybe a comment above just saying this is magic wand would help. I was expecting to see magic wand among the defs. (Or if not, where is magic wand, don't we need it?)

Kraken/Examples/Examples.lean
```diff
@@ -430,3 +407,223 @@ example [layout : Layout] (s: MachineData)
...
+    movq -8(%rsp), %rbx
+")
+
+theorem dynamic_stack_example_correct [layout : Layout] (s₀ : MachineData)
+    (stack : List UInt8) (lstack : stack.length = 1024) R
+    (h : s₀.regs.r9.toNat + s₀.regs.r15.toNat < 125)
+    (h_mem : s₀.dmem =⋆ stack.At (s₀.regs.rsp.toBitVec - 1024) ⋆ R) :
+    Eventually (straightlineStep (layout dynamic_stack_example))
+      (fun s' => s'.1.regs.rax = 42 ∧ s'.1.regs.rbx = 99 ∧ s'.1.regs.rsp = s₀.regs.rsp)
+      (s₀, layout.start) := by
```
aferr, line 556: Just to confirm, these do look like what I was hoping for for the first 3 (aside from us neeeding a complete ex3)

aferr:
> ```
> theorem BitVec.take_all {w : Nat} (x : BitVec w) : x.take w = x := sorry
> 
> ...
> 
> theorem UInt64.toBytes_ofBytes (bs : List UInt8) (h : 8 ≤ bs.length) : (UInt64.ofBytes bs).toBytes = bs.take 8 := sorry
> ```

Is this updated list of theorems your wishlist?

Kraken/SeparationPredicates.lean
andres-erbsen: No magic wand so far, this is just function application.

andres-erbsen:
The theorems in the top comment are meant as a summary of top-level stuff added.

Kraken/Mem.lean
```diff
@@ -0,0 +1,212 @@
...
+open Std.ExtHashMap
+open List
+
+def List.allSome {α} (l : List (Option α)) : Option (List α) := l.mapM id
+
+def List.At {w : Nat} (bs : List UInt8) (a : BitVec w) : ExtHashMap (BitVec w) UInt8 :=
+  .ofList (bs.mapIdx (fun i b => (a + BitVec.ofNat w i, b)))
+
+def UInt64.At {w : Nat} (val : UInt64) (a : BitVec w) : ExtHashMap (BitVec w) UInt8 :=
+  val.toBytes.At a
```
andres-erbsen: I figure we should either duplicate this for UInt* or remove it

Kraken/Examples/Examples.lean
```diff
@@ -430,3 +395,202 @@ example [layout : Layout] (s: MachineData)
...
+    bv_decide
+  rw [h_addr_eq] at h_mem
+  replace h_mem : ((stack.drop 1016).At (rsp.toBitVec + BitVec.ofNat 64 (2^64 - 8)) ⋆ ((stack.take 1016).At (rsp.toBitVec - 1024#64) ⋆ R)) _ := cast (congrFun (by ac_rfl) _) h_mem
+  have h_mem1 := Mem.storeInt_sep (rsp.toBitVec + BitVec.ofNat 64 (2^64 - 8)) 8 (stack.drop 1016) ((stack.take 1016).At (rsp.toBitVec - 1024#64) ⋆ R) mem ⟨h_mem, h_len_drop⟩ 99
+  rw [store_sep ss]
+  case h_mem => exact h_mem
+  case h_len => exact h_len_drop
+  sym =>
+  -- FIXME: kstep here takes too long
+  sorry
```
aferr, line 596: It looks like the time for kstep is a barrier here. If it simplifies things, I'm OK with opening an issue to track finalizing this example as a task to be solved at some later time. If we go that route, perhaps it's best to split this WIP proof body to a seprate branch and leave the body of this proof as just sorry to check in for this current PR? (That way we do not wind up needing to maintain the WIP proof in case some other update changes it).

Kraken/Examples/Examples.lean
```diff
@@ -430,3 +395,202 @@ example [layout : Layout] (s: MachineData)
...
   /- sorry -/
 
+
+open Std
+open Std.ExtHashMap
+
+theorem BitVec.take_all {w : Nat} (x : BitVec w) : x.take w = x := by
+  simp [BitVec.take]
+
+def move_2_regs_to_heap := parse("
```
aferr, line 405: Since Examples is growing (which is great!), can we consider splitting out these separation logic ones to a new file (perhaps SeparationExamples / MemoryExamples, possibly renaming Examples.lean to BasicExamples.lean or similar)?

Kraken/Mem.lean
```diff
@@ -0,0 +1,114 @@
...
+  .ofList (bs.mapIdx (fun i b => (a + .ofNat w i, b)))
+
+def Mem.storeBytes {w} (m : Mem w) (a : BitVec w) (bs : List UInt8) : Mem w :=
+  m.union (bs.At a)
+
+def Mem.storeInt {w} (m : Mem w) (a : BitVec w) (n : Nat) (v : Int) : Mem w :=
+  storeBytes m a (Int.toBytes n v)
+
+
+def UInt64.At {w} (val : UInt64) (a : BitVec w) : ExtHashMap (BitVec w) UInt8 :=
```
aferr: Should the return type here (and the same for the 3 other defs below) be `Mem (w)` or is there a reason it is not used here?

Kraken/Mem.lean
```diff
@@ -0,0 +1,114 @@
...
+def UInt32.At {w} (val : UInt32) (a : BitVec w) : ExtHashMap (BitVec w) UInt8 :=
+  val.toBytes.At a
+
+def UInt16.At {w} (val : UInt16) (a : BitVec w) : ExtHashMap (BitVec w) UInt8 :=
+  val.toBytes.At a
+
+def UInt8.At {w} (val : UInt8) (a : BitVec w) : ExtHashMap (BitVec w) UInt8 :=
+  val.toBytes.At a
+
+
```
aferr, line 52: IIRC there was a decision that we don't have any theorems globally decorated with`@simp`, and this is partly why none here are in the global simp set right? That decision sounds reasonable to me, I just thought I'd ask to make sure this is deliberate here.

Kraken/Semantics.lean
```diff
@@ -90,7 +88,7 @@ structure StatusFlags where
   of : Bool
   deriving Repr, BEq, DecidableEq, Hashable, Lean.ToExpr
 
-abbrev DataMem := Std.ExtHashMap UInt64 UInt64
+abbrev DataMem := Std.ExtHashMap (BitVec 64) UInt8
```
aferr: Why not ` DataMem := Mem 64` ?

Kraken/Separation/Lemmas.lean
aferr: I don't know what that standard is. It sounded like there was some intention to discuss our implementation of separation logic with Lean FRO members, so perhaps a note should be made to just ask them this convention or you might want to ask directly in the zulip about the convention.

Aside from not knowing their policy, I think omitting as you've done here is better because it makes the theorem more general.

Can you make it consistent within the file though? Does `union_comm_of_disjoint` need it for example (and if it does not please omit it to be consistent). Just do a quick trial to make sure the instances are omitted where possible?

Kraken/SeparationPredicates.lean
aferr: Is this because we want to treat wand as a separate task (ex to include any related proof rules along with it) or because you believe we do not need it? I would guess the former, but want to ask anyway. Can we open an issue to track this (even just a single short setence worth)? If you have an idea in mind of the remaining tasks to have our "complete" implementation of separation logic, it might be wortwhile to have an issue that is just a bullet point list enumerating these.

Kraken/Syntax.lean
aferr: I don't object.

Kraken/Theorems.lean
```diff
@@ -1,74 +0,0 @@
-/-
-Kraken - Helper Theorems
--/
-
-import Kraken.Semantics
-import Kraken.Syntax
-
--- UInt64.ofInt (k : Int) ≠ 0 when k is a natural number with k < 2^64 and k ≠ 0
```
aferr, line 8: I am fine with all 3 of these being deleted.

Kraken/Examples/Examples.lean
```diff
@@ -430,3 +395,202 @@ example [layout : Layout] (s: MachineData)
...
+  dsimp only [move_2_regs_to_heap]
+  rw [Executable.directivesFromStart]
+  simp [List.mapIdx, List.mapIdx.go]
+  sym => kstep; tactic =>
+  simp [AddrExpr.interp, ConstExpr.interp, Reg64s.get64, Width.bits, BitVec.toAddressSize, BitVec.signed, BitVec.take_all, BitVec.ofInt_toInt]
+  have h_mem1 := Mem.storeInt_sep rdi.toBitVec 8 v1.toBytes (v2.At (rdi.toBitVec + 8#64) ⋆ R) mem ⟨h_mem, h_bs1⟩ rax.toBitVec.toInt
+  rw [store_sep]
+  case h_mem => exact h_mem
+  case h_len => exact h_bs1
+  replace h_mem1 : (v2.At (rdi.toBitVec + 8#64) ⋆ ((Int.toBytes 8 rax.toBitVec.toInt).At rdi.toBitVec ⋆ R)) _ := cast (congrFun (by ac_rfl) _) h_mem1
```
aferr: `cast (congrFun (by ac_rfl) _)` comes up a few times, is it worth packing this into a tactic or other way of reducing the repeated boilerplate? (If the effort to do this is not so high)

Kraken/SeparationPredicates.lean
```diff
@@ -0,0 +1,14 @@
+import Std
+namespace Std.ExtHashMap
+local notation "map" => Std.ExtHashMap
+
+variable {key value : Type} [BEq key] [EquivBEq key] [Hashable key] [LawfulHashable key]
+
+instance : CoeFun (map key value) (fun _ => map key value → Prop) where coe := Eq
+
+def sep (p q : map key value → Prop) (m : map key value) : Prop :=
```
aferr: I suspect it is wortwhile to give heap predicates (`map key value -> Prop`) an explicit name both for clarity and in case these change in definition.

Kraken/SeparationPredicates.lean
```diff
@@ -0,0 +1,14 @@
...
+  ∃ a b, a.union b = m ∧ a.inter b = ∅ ∧ p a ∧ q b
```
aferr: Beyond wand some other core parts of the Separation Predicates seem missing,  namely:
* the empty heap (presumably built into ExtHashMap, but possibly worth an abbrev)?
* the singleton heap (mapping one var to one val and of size one)
* lifting a "pure" non-heap predicate into a heap predicate
* existential quantification

The software foundations chapter (https://softwarefoundations.cis.upenn.edu/slf-current/Hprop.html) also has distinct definition for separation between postconditions and heap predicates. Possibly we want this as well but we need to make this agree with the omnisemantics idea of postconditions, so fine to defer that one.

Can we define these in this PR even if we do not necessarily write all the proof rules / theorems that go along with them?

aferr: Perhaps with your design these can be relatively straightforward like so:

```
abbrev HeapProp := map key value → Prop

abbrev emp : HeapProp := ∅

abbrev singleton (addr : key) (val : value) : HeapProp :=
  .ofList [(addr, val)]

def pure (P : Prop) : HeapProp := fun m => P ∧ m = ∅

def hexists {α : Type} (F : α → HeapProp) : HeapProp :=
  fun m => ∃ x, F x m
```

(I suspect those are decent choices for where to do abbrev vs def, but open-minded!)

Kraken/SeparationPredicates.lean
```diff
@@ -0,0 +1,14 @@
+import Std
+namespace Std.ExtHashMap
+local notation "map" => Std.ExtHashMap
+
+variable {key value : Type} [BEq key] [EquivBEq key] [Hashable key] [LawfulHashable key]
+
+instance : CoeFun (map key value) (fun _ => map key value → Prop) where coe := Eq
```
aferr: This will apply globally to any file that imports this one. So hypothetically we could get behavior like this:
```
let cache : ExtHashMap String Nat := ...
let x := cache someOtherMap   -- silently becomes: cache = someOtherMap : Prop
```

Can you please compare the choice taken here to these two other options:

1) using `scoped instance` for this CoeFun instance and keeping the rest mostly the same. This would apply only in files that import this one and do `open Std.ExtHashMap` I think.

2) Giving an explicit and opaque def for HeapPreds and applying this instance just to the new type name like so:

```
def HeapPred (key value : Type) [BEq key] [Hashable key] := 
  ExtHashMap key value → Prop

instance : Coe (ExtHashMap key value) (HeapPred key value) where 
  coe m := (· = m)

instance : CoeFun (HeapPred key value) (fun _ => ExtHashMap key value → Prop) where 
  coe := id

def sep (p q : HeapPred key value) : HeapPred key value :=
  fun m => ∃ a b, a.union b = m ∧ a.inter b = ∅ ∧ p a ∧ q b
```

I lean towards this second choice (and I like having an explicit name for heap predicates even if they wind up being something else like an abbrev just for readability).

Kraken/Separation/Lemmas.lean
```diff
@@ -0,0 +1,155 @@
...
+    have ⟨a, bc, habc_union, habc_inter, hp, ⟨b, c, hbc_union, hbc_inter, hq, hr⟩⟩ := h
+    subst hbc_union
+    refine ⟨a.union b, c, (union_assoc a b c).trans habc_union, ?_, ⟨a, b, rfl, ?_, hp, hq⟩, hr⟩
+    · have ⟨hab, hac⟩ := (disjoint_union_r a b c).mp habc_inter
+      exact (disjoint_union_l a b c).mpr ⟨hac, hbc_inter⟩
+    · exact ((disjoint_union_r a b c).mp habc_inter).1
+
+instance : Std.Commutative (sep : (ExtHashMap key value → Prop) → _) := ⟨sep_comm⟩
+instance : Std.Associative (sep : (ExtHashMap key value → Prop) → _) := ⟨sep_assoc⟩
+
```
aferr, line 155: Missing identity laws ? 

```
theorem sep_emp (p : HeapPred key value) : p ⋆ emp = p
theorem emp_sep (p : HeapPred key value) : emp ⋆ p = p
```

Kraken/Separation/Lemmas.lean
```diff
@@ -0,0 +1,155 @@
+import Kraken.SeparationPredicates
+
```
aferr: Can we put the 3 `Separation*.lean` files into a new subfolder `Kraken/SeparationLogic` (or `Kraken/Separation` ?

Kraken/ToBytes.lean
aferr: very optionally: open an issue if you think someone might close it one day?

I am fine with this inclusion.

aferr (changes requested):
Thanks for starting this! This is a pretty important and valuable change! Looking forward to the followup.

Kraken/SeparationPredicates.lean
andres-erbsen: I don't think we need the magic wand for anything we might do with kraken. Other connectives may be useful for e..g linked data structures, but per Jonathan's suggestion I am deferring these too

Kraken/Examples/Examples.lean
andres-erbsen: What's the reason for splitting? @protz what would you prefer for hacking on `kstep`?

Kraken/SeparationPredicates.lean
andres-erbsen: I tried this, but doing so seems to break coercions. It would be neat to have it, and it would also help with the coercion issue you pointed out. I pushed the change anyway so you can take a look. I don't fully understand what's going on.

<details>
  <summary>Gemini's guess</summary>
  ### Why it worked before vs. now

  1. Before (Hardcoded  SepPred (w : Nat) ):
  Because  SepPred  explicitly required  w : Nat , checking  s₀.dmem : Mem 64  immediately forced  w = 64  on the
  right-hand side before checking  v1.At s₀.regs.rdi . Because Lean knew the expected argument type was precisely
  BitVec 64 , its target-directed coercion ( Coe UInt64 (BitVec 64) ) automatically fired and inserted  UInt64.
  toBitVec  for you.
  2. Now (Generic  SepPred (key value : Type) ):
  Because  SepPred  is now generic over  key  and  value , when  v1.At s₀.regs.rdi  ( ExtHashMap (BitVec ?w) UInt8 )
  is elaborated inside  p ⋆ q ,  ?w  is still an unconstrained metavariable ( ?w ).
  When Lean checks  s₀.regs.rdi : UInt64  against the expected parameter  a : BitVec ?w , it cannot trigger  Coe
  UInt64 (BitVec ?w)  because  ?w  is unknown during typeclass synthesis. Explicitly calling  .toBitVec  ( BitVec 64
  ) resolves  ?w = 64  immediately and satisfies the parameter type.

────────────────────────────────────────────────────────────
> isn't there already a coercion from Uint* to BitVec?

▸ Thought for 2s, 424 tokens
  The user asks: "isn't there already a coercion from Uint to BitVec?"
  Yes! There is already a built-in coercion ( Coe UInt64 (BitVec 64)  via  UInt64.toBitVec ).

  However, in Lean 4, any target-directed coercion  Coe α β  only fires when the expected target type  β  is fully
  known at the exact moment the argument is checked.

  ### Exactly why the built-in coercion didn't fire

  • When  SepPred  was hardcoded to  Mem w → Prop , the expected target type for  a  in  v1.At a  was already known
  to be precisely  BitVec 64 . Because  β = BitVec 64  was concrete, Lean synthesized  Coe UInt64 (BitVec 64)  and
  fired the built-in coercion automatically.
  • With generic  SepPred (key value : Type) , when Lean encounters  v1.At s₀.regs.rdi , the width  ?w  in  BitVec ?w
  has not yet been unified with  64 . When Lean checks  s₀.regs.rdi : UInt64  against the target  BitVec ?w , it
  attempts to synthesize  Coe UInt64 (BitVec ?w) . Because  ?w  is an unconstrained metavariable, typeclass
  resolution refuses to guess  ?w = 64  during  Coe  search, so the built-in coercion doesn't trigger.

  Explicitly writing  .toBitVec  forces  s₀.regs.rdi  to be  BitVec 64 , which immediately unifies  ?w = 64  (
  ExtHashMap (BitVec 64) UInt8 ), allowing  sep  and  =⋆  to proceed cleanly without metavariable ambiguity.
<details>

Kraken/SeparationPredicates.lean
andres-erbsen: I added them at first ([lemmas](https://github.com/AeneasVerif/kraken/commit/48c85d01169098197d3a010b2ffe150d6799a86f), [definitions](https://gist.github.com/andres-erbsen/08ae50e37c04f281c619497f91db770f) because github seems broken at displaying them) but after discussing with we decided to hold off until we actually have a use case for them. `Eq` subsumpes `singletion`, `hexists` would be needed for linked data structures with pointers, and I'm not sure we'd use `pure` given that we also have a logic outside separation.

Kraken/SeparationPredicates.lean
andres-erbsen: I share your concenrs, and I also think having a coercion for maps to arguments of sep is essential for  usability -- we will be writing *a bunch* of these. I don't know how to get best of both worlds here, but I pushed my latest attempt in case you can see a better way.

Kraken/Separation/Lemmas.lean
andres-erbsen: Also missing identity :) See above

Kraken/ToBytes.lean
andres-erbsen: https://github.com/AeneasVerif/kraken/issues/123

Kraken/Examples/Examples.lean
aferr: The motivation is just understandability / organization of the code base. There's no need for this organization change beyond this and I am not adamant about it. Making sure kstep can progress is also more important than organization at this phase, so let's see what protz says, yes.

I do think categorizing / splitting the examples we have as they grow will eventually be useful for both humans and agents to understand and find the right examples. But I can be convinced that this PR is not the right time for that.

Kraken/SeparationPredicates.lean
aferr: So if `hexists` has the use-case you described, doesn't that mean it should be added, or do I misunderstand?

Regarding `pure` -- yes we do have a logic outside separation but isn't this exactly how we lift from that logic into thise one about heaps / separation? How does one use the rest of the logic without this?

Kraken/SeparationPredicates.lean
aferr: Nice! I do think Coe is better than CoeFun! Thanks also for giving SepPred a name.

I believe `def SepPred` (instead of `abbrev`) should give you best of both worlds. I think then the `Coe` still fires wherever `SepPred` is explicitly expected (like in `sep`'s signature), proofs work unchanged since it's still definitionally transparent, but the coercion stops leaking to unrelated `ExtHashMap → Prop` contexts. Worth trying?

Kraken/SeparationPredicates.lean
aferr: Thanks for explaining on magic wand. There's another thread about the remaining connectives, so I'll resolve this one.

Kraken/Separation/Lemmas.lean
andres-erbsen: I proposed some renames, PTAL

Kraken/SeparationPredicates.lean
andres-erbsen: `hexists` - let's add it when we add some code that uses it, right now it would be dead code.

`pure` - we don't need to put logic inside separation. For exmaple, a precondition would be `n <= 10 /\ m =* P * Q`, where `P * Q` is separation logic and `n <= 10` is normal logic. `pure` becomes common when the option of using normal logic outside separation logic is not present.

aferr: re hexists -- OK

pure -- OK I see now it's not strictly needed to connect these.

Kraken/Separation/Lemmas.lean
aferr: Overall, I'm not going to be stickler for organization.  But I thought I'd give my thoughts on the new structure:

I see there are folders for Separation and Memory each with lemmas now -- introducing these folders to contain just one file each seems odd. Do you have a clear case for what else goes in these folders? Is the idea that eventually the set of Lemmas grows enough to warrent splitting? (It also looks odd to have Mem.lean outside the folder called Memory in the current state).

Kraken/Examples/Examples.lean
protz: let's clean up later if you don't mind -- I don't want to pile too many requests onto this PR. Generally I think it's reasonable to separate into minimal / memory / full-fledged examples.

Kraken/Examples/Examples.lean
protz: generally speaking, we need to improve automation because this tactic script is repetitive and could be most likely automated, perhaps by integrating more reasoning into kstep, or having more simp lemmas -- can we do a followup issue and/or PR to improve proof scripts?

Kraken/Separation/Lemmas.lean
protz: I reviewed the two Lemmas*.lean file and I could not infer the intent as to what kind of lemma goes into which file. I thought Separation/Lemmas.lean would have everything related to the * operator, while Memory/Lemmas.lean would have facts about ExtHashMap with no * operator, but there are counter-examples for both.

@andres-erbsen what was your guiding principle?

Kraken/Mem.lean
```diff
@@ -0,0 +1,114 @@
...
+def List.At {w} (bs : List UInt8) (a : BitVec w) : ExtHashMap (BitVec w) UInt8 :=
+  .ofList (bs.mapIdx (fun i b => (a + .ofNat w i, b)))
+
+def Mem.storeBytes {w} (m : Mem w) (a : BitVec w) (bs : List UInt8) : Mem w :=
+  m.union (bs.At a)
+
+def Mem.storeInt {w} (m : Mem w) (a : BitVec w) (n : Nat) (v : Int) : Mem w :=
+  storeBytes m a (Int.toBytes n v)
```
protz, line 37: can we make it toBytesLe?

Kraken/Mem.lean
protz: we can and should have our own simpset -- maybe a followup for better proof automation?

Kraken/Mem.lean
```diff
@@ -0,0 +1,114 @@
...
+  val.toBytes.At a
+
+def UInt16.At {w} (val : UInt16) (a : BitVec w) : ExtHashMap (BitVec w) UInt8 :=
+  val.toBytes.At a
+
+def UInt8.At {w} (val : UInt8) (a : BitVec w) : ExtHashMap (BitVec w) UInt8 :=
+  val.toBytes.At a
+
+
+theorem mem_At_iff {w} (bs : List UInt8) (a : BitVec w) (k : BitVec w) :
```
protz, line 53: I'm having a hard time making sense of why lemmas are here too

Kraken/Separation.lean
```diff
@@ -0,0 +1,17 @@
+import Std
+
+namespace Std.ExtHashMap
+local notation "map" => Std.ExtHashMap
```
protz: where is this notation used?

Kraken/Separation.lean
```diff
@@ -0,0 +1,17 @@
...
+
+abbrev SepPred (key value : Type) [BEq key] [Hashable key] := ExtHashMap key value → Prop
+
+variable {key value : Type} [BEq key] [EquivBEq key] [Hashable key] [LawfulHashable key]
+
+instance : Coe (ExtHashMap key value) (SepPred key value) where coe := Eq
+
+def sep (p q : SepPred key value) (m : ExtHashMap key value) : Prop :=
```
protz: nitpick: `def sep p q: SepdPrev k v := fun m => ...` (to leverage the type annotation)

Kraken/ToBytes.lean
```diff
@@ -0,0 +1,383 @@
+abbrev Nat.take (v : Nat) (w : Nat) : Nat := v % 2 ^ w
+abbrev Int.take (v : Int) (w : Nat) : Int := v % (2 : Int) ^ w
+
+instance : Coe UInt8 (BitVec 8) := ⟨UInt8.toBitVec⟩
+instance : Coe UInt16 (BitVec 16) := ⟨UInt16.toBitVec⟩
+instance : Coe UInt32 (BitVec 32) := ⟨UInt32.toBitVec⟩
+instance : Coe UInt64 (BitVec 64) := ⟨UInt64.toBitVec⟩
+
+def Int.ofBytes (bs : List UInt8) : Int :=
```
protz, line 9: let's suffix with Le or Be to make things clear

protz (commented):
I will look at this in more detail and especially the part about coercions

Kraken/Mem.lean
protz: Andres' response: "let's pretend be never existed"

Kraken/SeparationPredicates.lean
andres-erbsen: `def` doesn't seem to make a difference in normal tactics, and while I didn't try it in SymM I'd expect it to break even more typechecking. I pushed a version that avoids coercions and uses Eq

Kraken/ToBytes.lean
andres-erbsen: let's pretend big-endian never existed, we can write `Int.ofBytes (List.rev ...)` if we need it

protz:
Ok, I'm about to hit the merge button on this. Recap of the discussions:
- we leave automatic coercions out of the equation for now, and once we have a larger body of code, we'll be able to assess whether Coe or CoeFun are acceptable and don't pollute client code
- lemmas are either in Mem (base memory), SeparationMem (separation view of the memory), Separation (separation logic and associated properties)
- whatever helper lemmas are required initially are marked private and live in the same file as the main lemma; should they be frequently needed, they can be upgraded to StdLibCandidates as a pathway for inclusion in Lean itself

@andres-erbsen I summarized our discussion from yesterday into DEVELOPER.md -- feel free to submit a followup PR if need be

PR #121, opened 2026-06-17
Move Fixes (with tests against GNU as):

aferr:
These are the fixes
1) The parser incorrectly mapped suffix-sized sign-extension instructions to the zero extension versions of these.
2) There appeared to be a typo mapping boszx to movsx. 
3) Missing parseCommas for movsx and movzx

The tests should fail if you checkout Parser.lean from main and run these tests against them.

### DIFF
--- a/Kraken/Parser.lean
+++ b/Kraken/Parser.lean
@@ -621,15 +621,15 @@ def parseInstr : Parser Instr := do
     let w ← instrWidth mn
     commaSeparated w parseOperand parseRegOrMem .mov
 
-  | "moszx" =>
+  | "movsx" =>
     -- Must be a register otherwise lacking type info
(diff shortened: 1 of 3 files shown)


andres-erbsen:
What's the overlap between this change and https://github.com/AeneasVerif/kraken/pull/88?

aferr:
> What's the overlap between this change and #88?

Ah looks like #88 covers this so this PR can be closed. I didn't notice 88. Might be good to get that in so you don't keep getting fixes for it.

aferr:
@andres-erbsen actually unless you think #88 is further out since it seems to be working on a broader feature? Could be worth splitting out the bug fix part and getting that in if the rest of that PR is farther from landing ?

PR #125, opened 2026-07-21
Add initial support for SSE/AVX registers ({x,y,z}mm)s, {v,}movups

zrlk:
This PR adds ISA state support for the AVX registers zmm0-31, which alias to ymm0-31 and xmm0-31. It provides initial implementations for the {v,}movups instructions, leaving out writemasks for now.

There was (and still is) some question about how operand widths should be factored (i.e., if AVX widths should be part of `Width` or some special `AvxWidth`), in part because it permits us to represent some impossible registers (like a 512-bit rax). After starting to push through the PR with a separate width type (and just before adding AVX-specific widths to AVX insns), it seemed like going all the way through would result in a lot of duplication in Semantics (given that widths are pretty pervasive there). Furthermore, we already permit impossible instructions to be encoded (like insns that take two memory operands), so the infelicity of a very wide rax might be tolerable?

This is a draft CL. We don't currently test the ymm/zmms or all combinations of operator kinds, nor do we actually check the status of the xmm registers in tests (the test in the PR does this via the stack). Other deficiencies are expected.

### DIFF
--- a/Kraken/Parser.lean
+++ b/Kraken/Parser.lean
@@ -137,6 +137,11 @@ def parseInt : Parser Int := do
   let val ← parseHexOrDec
   pure (if neg then -val else val)
 
+/-- Helper to parse a standard natural number. -/
+def parseNat : Parser Nat := do
+  let digits ← many1 digit
+  pure (digits.foldl (fun acc d => acc * 10 + (d.toNat - '0'.toNat)) 0)
+
 /-- Parse a name (identifier or label). -/
 def parseName : Parser String := do
   let first ← satisfy fun c => c.isAlpha || c == '_' || c == '.'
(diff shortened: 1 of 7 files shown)


Kraken/Parser.lean
```diff
@@ -137,6 +137,11 @@ def parseInt : Parser Int := do
   let val ← parseHexOrDec
   pure (if neg then -val else val)
 
+/-- Helper to parse a standard natural number. -/
+def parseNat : Parser Nat := do
```
protz: there is`digits` in Lean's parser library, but perhaps it doesn't work here?

zrlk: Yes, `digits` works here. I removed parseNat.

Kraken/Parser.lean
```diff
@@ -625,34 +709,34 @@ def parseInstr : Parser Instr := do
...
     let src ← parseRegO w_src; parseComma
     let dst ← parseRegO w_dst
-    pure ⟨ .W64, w_dst, .movzx dst (.reg src) ⟩
+    pure (.regular .W64 w_dst (.movzx dst (.reg src)))
 
   | "lea" =>
     let (addr_w, src) ← parseMemory; parseComma
     let ⟨ w, dst ⟩ ← parseRegW
-    pure ⟨ addr_w, w, .lea dst src ⟩
+    pure (.regular addr_w w (.lea dst src))
```
protz: do we want coercions here to cast regular and avx operations into the instruction type?

protz (approved):
Looks good -- I like that this has almost no impact on the proofs!

Kraken/Parser.lean
zrlk: This cleaned up the parser a little bit.

PR #126, opened 2026-07-23
fix: avoid ZF, CF, PF disagreements by fixing a starting stack

zrlk:
We need to start the stack at a plausible address, otherwise we get trivial CF/ZF flag disagreements. Since we don't control the address of the stack when running natively (but we *can* control the last byte), byte-align the stack to avoid PF disagreements.

### DIFF
--- a/Kraken/Test/asm/test_pop_mem_rsp.S
+++ b/Kraken/Test/asm/test_pop_mem_rsp.S
@@ -11,5 +11,6 @@ sub %rsi, %rdi
 sub %rsi, %rax
 mov %rsi, %rsp
 
-xor %rsi, %rsi
-xor %r11, %r11
+# AF is undefined after xor
+mov $0, %rsi
+mov $0, %r11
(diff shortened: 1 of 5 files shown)


Kraken/Test/asm/test_pop_mem_rsp.S
```diff
@@ -11,5 +11,6 @@ sub %rsi, %rdi
 sub %rsi, %rax
 mov %rsi, %rsp
 
-xor %rsi, %rsi
-xor %r11, %r11
+# AF is undefined after xor
```
protz: FYI, we also have a facility for stating that a test leaves some flags undefined and not checking for those in the expected test output (syntax is: `# Undefined flags: ...`, used in several other tests)

protz (approved):

protz:
Thanks for the great comments!

Kraken/Test/asm/test_pop_mem_rsp.S
zrlk: I think I got those in along with the comments in https://github.com/AeneasVerif/kraken/pull/126/changes/d4c5844394825d065105283b9225307c6598a80c -- I'll make sure to use them going forward.

protz: yeah sorry this was a stale review comment from before we decided to talk about this over coffee

PR #128, opened 2026-07-28
Initial AArch64 support

lukaszobernig:
This adds initial support for AArch64: syntax, semantics, assembly parser, a QEMU based test runner and tests.

### DIFF
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -14,9 +14,19 @@ jobs:
       - uses: leanprover/lean-action@v1
 
   run-asm-tests:
-    name: Run ASM tests (Ubuntu)
+    name: Run X64 ASM tests (Ubuntu)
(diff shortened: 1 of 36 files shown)


protz (changes requested):
Overall this looks great. I particularly appreciated your comments in Syntax.lean as I'm not familiar with ARM assembly syntax. Some general thoughts:
- I think some sharing of definitions would help; perhaps not everything is shareable, but at least parsing helpers, machine load helpers, and so on, seem like they can be. This might unlock further sharing of lemmas especially related to separation.
- Same question with the test runner -- some of the lean dylib machinery can perhaps be shared?

protz (approved):
We discussed this live -- there are not many opportunities for cleanup beyond trivial helpers at the beginning of syntax, semantics and parser.

Let's merge this and then followup up with a cleanup PR.

PR #129, opened 2026-07-30
refactor: golfing of various sorts, introduce the kprologue tactic

SamuelSchlesinger:
my attempt at some helpful cleanup


protz (approved):
That's a great PR -- thanks for the cleanup and writing that `kprologue` tactic

PR #130, opened 2026-07-31
fix: trace instead of log in klog so we can keep it disabled by default

SamuelSchlesinger:
calling it `klog` still is probably a faux pas, but this is quite noisy otherwise


protz (approved):

protz:
thanks -- wasn't sure what was idiomatic here; feel free to suggest better names, although with my branch protz/tactics under heavy development, probably best to leave renames to until protz/tactics is merged

SamuelSchlesinger:
I mean, `ktrace` is a fine change! I only meant to change it now that it is tracing and not merely logging

PR #131, opened 2026-07-31
feat(x64): compare ymm register state and sparsely produce output

zrlk:
This PR adds support for actually checking the values of ymm0-ymm15 against the Kraken semantics. (It seems that most of our machines don't support the extensions required for the zmms, or for *mms above 15).

This PR will also only populate GPRs in the Kraken result json when they have nonzero values. This should reduce noise.

With this change, krakenrunner now produces output that looks like:

```json
{
  "flags":{"af":false,"cf":false,"of":false,"pf":false,"sf":false,"zf":false},
  "regs":{"r10":10,"r11":11,"r12":12,"r13":13,"rcx":44,"rdi":42,"rdx":44,"rsi":43},
  "zmms": {
    "zmm0": "2a000000000000002b",
    "zmm1": "2c000000000000002d",
    "zmm2": "a000000000000000b000000000000000c000000000000000d"
  }
}
```

(on test_avx_mov.S).

I haven't taken a stance about how we should stringify the values of the zmm registers; for now I just did the most direct thing. We may want to split these into vector lanes in the future (eg we'd print ymm2 as <13,12,11,10> instead).

### DIFF
--- a/Kraken/X64/Test/asm_tests.py
+++ b/Kraken/X64/Test/asm_tests.py
@@ -160,7 +160,7 @@ def compare_states(real: ExecutionState, kraken: ExecutionState, undefined_flags
         if rv != kv:
             diffs.append(f"{r}: x86={rv:#x} ({rv}), kraken={kv:#x} ({kv})")
 
-    for r in [r for r in ZMMS]:
+    for r in ZMMS:
         rv, kv = real.zmms.get(r, '0'), kraken.zmms.get(r, '0')
         if rv != kv:
             diffs.append(f"{r}: x86={rv}, kraken={kv}")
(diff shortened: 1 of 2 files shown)


Kraken/X64/Test/asm_tests.py
```diff
@@ -141,10 +156,15 @@ def get_undefined_flags(path: Path) -> List[str]:
 def compare_states(real: ExecutionState, kraken: ExecutionState, undefined_flags: List[str]) -> List[str]:
     diffs = []
     for r in [r for r in REGS if r != "rsp"]:
-        rv, kv = real.regs[r], kraken.regs[r]
+        rv, kv = real.regs.get(r, 0), kraken.regs.get(r, 0)
         if rv != kv:
             diffs.append(f"{r}: x86={rv:#x} ({rv}), kraken={kv:#x} ({kv})")
 
+    for r in [r for r in ZMMS]:
```
protz: nit: how does this differ from `for r in ZMMS:`?

protz (approved):
Looks good! I only have one minor nit, but feel free to merge anyhow

Kraken/X64/Test/asm_tests.py
zrlk: It's less efficient for no reason! Fixed.

PR #132, opened 2026-08-03
Update Aarch64 support

lukaszobernig:
This adds some missing instructions, adds more comments to the parser and syntax, and restricts the parser to better align it with official assembler behaviour.

### DIFF
--- a/Kraken/AArch64/Semantics.lean
+++ b/Kraken/AArch64/Semantics.lean
@@ -525,18 +525,33 @@ def Operation.interp [Labels]
     let prod := val1.toNat * val2.toNat
     let res := BitVec.ofNat 64 (prod >>> 64)
     s.setRegOrZr dst res next
+  | .AND_i dst src1 imm =>
+    let val1 := s.regs.getRegOrZr src1
+    let val2 := BitVec.ofInt w.bits (Int64.toInt (imm.interp p))
+    let res := val1 &&& val2
+    s.setRegOrSp dst res next
   | .AND_s dst src1 src2 =>
     let val1 := s.regs.getRegOrZr src1
     let val2 := src2.interp s.regs p
     let res := val1 &&& val2
     s.setRegOrZr dst res next
+  | .ANDS_i dst src1 imm =>
+    let val1 := s.regs.getRegOrZr src1
+    let val2 := BitVec.ofInt w.bits (Int64.toInt (imm.interp p))
+    let res := val1 &&& val2
+    let flags := StatusFlags.from_result res { c := false, v := false }
+    { s with status := flags }.setRegOrZr dst res next
   | .ANDS_s dst src1 src2 =>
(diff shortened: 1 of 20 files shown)


Kraken/AArch64/Semantics.lean
```diff
@@ -313,6 +314,25 @@ def AddrExpr.interp [Labels] {w} (mem : AddrExpr w) (s : MachineData) (p : Std.R
...
+def UnscaledAddrExpr.eval [Labels] (mem : UnscaledAddrExpr) (s : MachineData) (p : Std.Rco Int64) : BitVec 64 :=
+  let base := (s.regs.getRegOrSp mem.base).toInt
+  let off := Int64.toInt (mem.imm.interp p)
+  BitVec.ofInt 64 (base + off)
+
+def UnscaledAddrExpr.checkSPAlignment (mem : UnscaledAddrExpr) (s : MachineData) (ok : Unit → Effects) : Effects :=
+  match mem.base with
+  | .SP =>
+    if s.regs.getRegOrSp .SP % 16#64 != 0#64 then
+      .unimplemented s!"Unimplemented: SP is required to be 16-byte aligned"
```
protz, line 326: is this unimplemented or an error case that we should add?

Kraken/AArch64/Semantics.lean
protz, line 418: but for sequential code I presume this is irrelevant?

protz (approved):
Thanks -- this is good and I appreciated the comments.

I only have minor thoughts that can be addressed in a followup.

Out of curiosity, how were your tests generated?

Kraken/AArch64/Semantics.lean
lukaszobernig: I'm using `unimplemented` as a temporary placeholder, there should really be a separate `sp_alignment` effect for this. I hadn't properly made my mind up, but I just saw that https://github.com/AeneasVerif/kraken/pull/133 is adding a similar alignment check for memory.

Kraken/AArch64/Semantics.lean
lukaszobernig: This should indeed be irrelevant for sequential code since we reject unpredictable cases now in the parser.

lukaszobernig:
> Out of curiosity, how were your tests generated?

For the parser tests I have tried to cover all possible cases for the different assembly constructs: correct parse, pseudo instructions, parser errors. I need to figure out how to generate test coverage to see if the current tests actually cover everything. I have also ran an AI to check if there are any missed cases (with mixed success).

For the runner tests, I have tried to cover all possible instruction forms and widths (32-bit, 64-bit) for each class. We can test many forms at once since we have so many output registers to compare to. But again, it would be good to double check the coverage.

Miriam and I discussed about perhaps using a fuzzer approach to generate tests with better coverage. I will have to look into that.

PR #133, opened 2026-08-03
feat(x64): movaps, subps, addps

zrlk:
Implement more legacy SSE insns with varying alignment requirements;
implement alignment checks as effects.

It is a little confusing to work out the tolerance (eg) ADDPS has for
unaligned m128s. The Intel SDM page for ADDPS only lists Table 2-19 ("Type
2 Class Exception Conditions") as a reference for VEX-encoded insns (which
legacy ADDPS is not), but section 2.5 ("Intel AVX and Intel SSE
Instruction Exception Classification"), which actually uses ADDPS as an
example, reads that ADDPS 'contains the entry "See Exceptions Type 2"',
leading with 'To look up the exceptions of legacy 128-bit SIMD
instructions...'. This shows that ADDPS raises #GP(0) for unaligned
addresses. (Curiously, the AVX variants do *not* require alignment.)

Note that ADDPS/SUBPS do not allow m128 dest ops.

### DIFF
--- a/Kraken/X64/Semantics.lean
+++ b/Kraken/X64/Semantics.lean
@@ -272,10 +272,19 @@ def MachineData.load
     | .some i => ret (.ofInt _ i) s
     | .none => nonmem_load s.dmem addr w (fun v dmem => ret v { s with dmem }))
 
+-- Alternatively, we could define this in terms of BitVecs without %:
+-- (addr &&& BitVec.ofNat 64 (bytes - 1)) == 0#64
+def isAligned (bytes : Nat) (addr : BitVec 64) : Bool :=
+  addr.toNat % bytes == 0
+
+-- Legacy SSE instructions are generally stricter about alignment requirements,
+-- while AVX (VEX-encoded) instructions can mostly deal with unaligned
+-- addresses (https://discourse.llvm.org/t/memory-alignment-model-on-avx-avx2-and-avx-512-targets/34705).
+-- For this reason we default checkAlign to false.
 def MachineData.loadAvx
   (s : MachineData) (addr : BitVec 64) (w : AvxWidth)
   (ret : w.type → MachineData → Effects) (checkAlign : Bool := false) : Effects :=
-  if checkAlign && (addr &&& BitVec.ofNat 64 (w.bytes - 1)) != 0#64 then
+  if checkAlign && !(isAligned w.bytes addr) then
     .gp_unaligned addr w.bytes
   else
     require_read_access addr .W64 (fun _unit =>
@@ -291,7 +300,7 @@ def MachineData.store (s : MachineData) (addr : BitVec 64) {w : Width} (v : w.ty
     | .none => nonmem_store s.dmem addr v (fun dmem' => ret { s with dmem := dmem' }))
 
 def MachineData.storeAvx (s : MachineData) (addr : BitVec 64) {w : AvxWidth} (v : w.type) (ret: MachineData → Effects) (checkAlign : Bool := false) : Effects :=
-  if checkAlign && (addr &&& BitVec.ofNat 64 (w.bytes - 1)) != 0#64 then
+  if checkAlign && !(isAligned w.bytes addr) then
     .gp_unaligned addr w.bytes
   else
     require_write_access addr .W64 (fun _unit =>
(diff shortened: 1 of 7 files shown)


Kraken/X64/Semantics.lean
```diff
@@ -169,6 +169,31 @@ def RegZmms.setLegacy (s : RegZmms) {w} (r : AvxReg w) (v : w.type) : RegZmms :=
 def BitVec.toAddressSize [address_size: AddressSize] (w: BitVec 64): BitVec address_size.address_size.bits :=
   w.take address_size.address_size.bits
 
+def BitVec.packedBinOp {w : Nat} (c : Nat) (op : BitVec c → BitVec c → BitVec c) (a b : BitVec w) : BitVec w :=
```
protz, line 174: a thought: would it be better to have a `split` helper that allows switching between e.g. u128 and 4xu32, and so on? I know that this is a common abstraction -- but perhaps we can add later on top of the recursively-defined function

Kraken/X64/Semantics.lean
```diff
@@ -248,11 +274,14 @@ def MachineData.load
 
 def MachineData.loadAvx
   (s : MachineData) (addr : BitVec 64) (w : AvxWidth)
-  (ret : w.type → MachineData → Effects): Effects :=
-  require_read_access addr .W64 (fun _unit =>
-match Mem.loadInt s.dmem addr w.bytes with
-    | .some i => ret (.ofInt _ i) s
-    | .none => unimplemented "AVX nonmem load not supported")
+  (ret : w.type → MachineData → Effects) (checkAlign : Bool := false) : Effects :=
```
protz, line 288: maybe add a comment here, because I thought the simd instructions required alignment by default, and checkAlign := false by default

Kraken/X64/Semantics.lean
```diff
@@ -248,11 +274,14 @@ def MachineData.load
...
+  if checkAlign && (addr &&& BitVec.ofNat 64 (w.bytes - 1)) != 0#64 then
```
protz: out of curiosity: I would have done a modulo here; do you foresee any advantage to stating this in terms of bitvecs?

protz: my assumption is that we'll carry alignment prerequisites through modulo

Kraken/X64/Semantics.lean
```diff
@@ -261,12 +290,15 @@ def MachineData.store (s : MachineData) (addr : BitVec 64) {w : Width} (v : w.ty
...
     | .none => nonmem_store s.dmem addr v (fun dmem' => ret { s with dmem := dmem' }))
 
-def MachineData.storeAvx (s : MachineData) (addr : BitVec 64) {w : AvxWidth} (v : w.type) (ret: MachineData → Effects) : Effects :=
-  require_write_access addr .W64 (fun _unit =>
-match Mem.loadInt s.dmem addr w.bytes with
-    | .some _ =>
-        ret { s with dmem := Mem.storeInt s.dmem addr w.bytes v.toInt }
-    | .none => unimplemented "AVX nonmem store not supported")
+def MachineData.storeAvx (s : MachineData) (addr : BitVec 64) {w : AvxWidth} (v : w.type) (ret: MachineData → Effects) (checkAlign : Bool := false) : Effects :=
+  if checkAlign && (addr &&& BitVec.ofNat 64 (w.bytes - 1)) != 0#64 then
```
protz: in any case I would define a helper `isAligned w addr`

protz (approved):
Looks good to me modulo my minor comments -- feel free to adjust & merge, or just merge directly.

Kraken/X64/Semantics.lean
zrlk: The oldest (most common at this point, I guess) SSE instructions (16-byte registers) are stricter, but the newer AVX variants (32- and 64-byte registers) are more forgiving: https://discourse.llvm.org/t/memory-alignment-model-on-avx-avx2-and-avx-512-targets/34705

We could possibly use a different default for the AvxLegacy variants (since the SSE instructions *also* differ in behavior wrt zeroing the top bits), but I think that might get confusing.

I've added a comment before the first checkAlign.

Kraken/X64/Semantics.lean
zrlk: I'm seeing some suggestion that `bv_decide` would work better with the bitvec version, but we can easily swap it back now that it's been pulled out to `isAligned`. Defining it in terms of % is definitely more readable.

Kraken/X64/Semantics.lean
zrlk: Done.

Kraken/X64/Semantics.lean
zrlk: I don't have great intuition about what will work best, especially when we start trying to write proofs that care about the actual values of the *MM registers (and not just addresses or basic load/store ops). For now I'll leave a TODO, but I want to come back to this soon.

PR #134, opened 2026-08-05
feat(x64): add l, le condition codes

zrlk:


Kraken/X64/Semantics.lean
```diff
@@ -344,6 +344,7 @@ match o with
 def CondCode.interp (cc : CondCode) (s : StatusFlags) : Bool := match cc with
   | .z  => s.zf | .nz => !s.zf | .c  => s.cf | .nc => !s.cf
   | .a  => !s.cf && !s.zf | .be => s.cf || s.zf
+  | .l => s.sf ^^ s.of | .le => (s.sf ^^ s.of) || s.zf
```
protz: I have a preference for sticking to the boolean format, because:
- it's consistent with the definitions above
- it follows the documentation
- it's easier to parse 

for instance, I'm reading this:

> 7F cb | JNLE rel8 | D | Valid | Valid | Jump short if not less or equal (ZF=0 and SF=OF).
> 7D cb | JNL rel8 | D | Valid | Valid | Jump short if not less (SF=OF).

so I would prefer `s.sf = s.of` and `s.zf = 0 && s.sf = s.of`

right now it requires a little bit of gymnastics to convince myself that what you wrote matches the doc... thanks!

protz (changes requested):
just a minor style comment, then let's merge this

Kraken/X64/Semantics.lean
zrlk: Good point, it's better to match the manual where we can. Done.

protz (approved):

protz:
Thank you!

PR #135, opened 2026-08-05
feat(x64): pretty-print directive lists/layouts

zrlk:
Machine-generated, take with a big grain of salt.

Turns
```lean
layout : Layout
s : MachineData
⊢ Effects.All
  (fun mid =>
    Eventually
      (straightlineStep
        (layout.apply
          [Directive.label "start",
            Directive.instr (Instr.regular Width.W64 Width.W64 (Operation.mov ↑(Reg.low Reg64.rax Width.W64) ↑↑1)),
            Directive.instr
              (Instr.regular Width.W64 Width.W64
                (Operation.xor ↑(Reg.low Reg64.rax Width.W64) ↑↑(Reg.low Reg64.rax Width.W64))),
            Directive.instr (Instr.regular Width.W64 Width.W64 (Operation.jcc CondCode.nz "start")),
            Directive.instr (Instr.regular Width.W64 Width.W64 (Operation.mov ↑(Reg.low Reg64.rax Width.W64) ↑↑2))]))
      (fun s => s.fst.regs.rax = 2) mid)
  (Directives.interp
    [(Directive.label "start", Layout.size 0),
      (Directive.instr (Instr.regular Width.W64 Width.W64 (Operation.mov ↑(Reg.low Reg64.rax Width.W64) ↑↑1)),
        Layout.size 1),
      (Directive.instr
          (Instr.regular Width.W64 Width.W64
            (Operation.xor ↑(Reg.low Reg64.rax Width.W64) ↑↑(Reg.low Reg64.rax Width.W64))),
        Layout.size 2),
      (Directive.instr (Instr.regular Width.W64 Width.W64 (Operation.jcc CondCode.nz "start")), Layout.size 3),
      (Directive.instr (Instr.regular Width.W64 Width.W64 (Operation.mov ↑(Reg.low Reg64.rax Width.W64) ↑↑2)),
        Layout.size 4)]
    s Layout.start fun pc s => Effects.done (s, pc))
```

into

```lean
layout : Layout
s : MachineData
⊢ Effects.All
  (fun mid =>
    Eventually
      (straightlineStep
        (layout.apply
          [asm|
              "start:"
              "mov rax, 1"
              "xor rax, rax"
              "jne start"
              "mov rax, 2"]))
      (fun s => s.fst.regs.rax = 2) mid)
  (Directives.interp
    [asm_layout|
        "@0: start:"
        "@1: mov rax, 1"
        "@2: xor rax, rax"
        "@3: jne start"
        "@4: mov rax, 2"]
    s Layout.start fun pc s => Effects.done (s, pc))
```


Kraken/X64/Tactics.lean
```diff
@@ -283,3 +285,178 @@ def evalSymKStep : Grind.GrindTactic :=
...
+
+-- Standard fallback syntax for general open lists
+syntax "[asm| " term,* " ]" : term
+
+--------------------------------------------------------------------------------
+-- 2. Define safe helpers to identify types & navigate arrays
+--------------------------------------------------------------------------------
+
+-- Fully compiler-independent indexing helper using core GetElem
+def getArg (args : Array Expr) (i : Nat) : Expr :=
```
protz, line 305: what is the point of having a default value rather than panicking or failing if this is an out-of-bounds access?

Kraken/X64/Tactics.lean
```diff
@@ -283,3 +285,178 @@ def evalSymKStep : Grind.GrindTactic :=
...
+  else
+    Expr.bvar 0
+
+def isDirectiveType (e : Expr) : MetaM Bool :=
+  return Expr.isConstOf e ``Directive
+
+def isNatType (e : Expr) : MetaM Bool :=
+  return Expr.isConstOf e ``Nat
+
+def isDirectiveNatType (e : Expr) : MetaM Bool :=
```
protz, line 317: how about `match_expr e with Directive \x Nat`? would that work?

Kraken/X64/Tactics.lean
```diff
@@ -283,3 +285,178 @@ def evalSymKStep : Grind.GrindTactic :=
...
+  | Expr.app (Expr.app (Expr.app (Expr.const ``OfNat.ofNat _) _) (Expr.lit (Literal.natVal n))) _ => some n
+  | _ => none
+
+--------------------------------------------------------------------------------
+-- 3. Define the consolidated Delaborator (Unsafe, getArg & Explicit Namespaces)
+--------------------------------------------------------------------------------
+
+-- Delaborates individual (Directive × Nat) pairs beautifully as "@N: instruction"
+@[app_delab Prod.mk]
+unsafe def delabDirectiveNatPair : Delab := do
```
protz, line 339: why do we need `unsafe` here?

Kraken/X64/Tactics.lean
```diff
@@ -283,3 +285,178 @@ def evalSymKStep : Grind.GrindTactic :=
...
+  let e ← getExpr
+  let args := e.getAppArgs
```
protz, line 341: is there better monadic syntax like `let #[ alpha, beta, .. ] := e.getAppArgs | failure`?

protz: or better, `match_expr`

Kraken/X64/Tactics.lean
```diff
@@ -283,3 +285,178 @@ def evalSymKStep : Grind.GrindTactic :=
...
+      let dExpr := getArg args 2
+      let nExpr := getArg args 3
+
+      -- 1. Evaluates closed Directive parts to their authentic ToString assembly format
+      let dStr ← if !dExpr.hasFVar && !dExpr.hasMVar then
+        let typeExpr ← Meta.inferType dExpr
+        let d ← Meta.evalExpr Directive typeExpr dExpr
+        pure s!"{d}"
+      else
+        pure "<symbolic_directive>"
```
protz, line 355: can we print something more useful here? like, if the directive is let-bound, I assume we'd want to know what's in there, rather than an opaque notation

Kraken/X64/Tactics.lean
```diff
@@ -283,3 +285,178 @@ def evalSymKStep : Grind.GrindTactic :=
...
+
+      -- 2. Extract the Nat index/offset from nExpr using fully-qualified matches
+      let idxStr ← match nExpr with
+        | Expr.app _ arg =>
+          if let some n := getNatVal? arg then
```
protz, line 360: I thought getNatVal? already looked under an `app`?

Kraken/X64/Tactics.lean
```diff
@@ -283,3 +285,178 @@ def evalSymKStep : Grind.GrindTactic :=
...
+            let idxStx ← withAppArg (withAppArg delab)
+            let fmt ← PrettyPrinter.ppTerm idxStx
+            pure s!"{fmt}"
+        | _ =>
+          if let some n := getNatVal? nExpr then
+            pure s!"{n}"
+          else
+            let idxStx ← withAppArg delab
+            let fmt ← PrettyPrinter.ppTerm idxStx
+            pure s!"{fmt}"
```
protz, line 373: can we differentiate concrete nats from symbolic layout nats?

Kraken/X64/Tactics.lean
```diff
@@ -283,3 +285,178 @@ def evalSymKStep : Grind.GrindTactic :=
...
+
+      let combinedStr := s!"@{idxStr}: {dStr}"
+      return Syntax.mkStrLit combinedStr
+  failure
+
+-- Recursive helper to process Lists of symbolic pairs
+unsafe def delabDirectiveNatListGo : DelabM (List Term) := do
+  let curr ← getExpr
```
protz, line 381: also something that might benefit from match_expr, if applicable

protz (changes requested):
Overall this looks good, I think we need to work a little harder to clean up and use helpers like `match_expr` everywhere possible. I'd like to take a second look post-cleanup. Thanks!

zrlk:
Thanks! I'm glad this didn't turn out to be *so* bad--as I handwaved in the intro comment, this was the result of a couple LLM iterations, so I'm not surprised that there's some basic weirdness. I'll get it into landable shape and let you know when it's ready for re-review.

Kraken/X64/Tactics.lean
zrlk: Failing makes sense here; done.

Kraken/X64/Tactics.lean
zrlk: I think we can do better here, yeah.

Kraken/X64/Tactics.lean
zrlk: Looks like there's another layer to get through (doing the naive thing here ended up giving me output like `@Kraken.Layout.size Directive 1: mov rax, 1` where we just want `@1: mov rax, 1`)

Kraken/X64/Tactics.lean
zrlk: Added a # in the format string for concrete nats.

Kraken/X64/Tactics.lean
zrlk: I don't think that will desugar here. I was able to clean up some of the isConstOf stuff though.

Kraken/X64/Tactics.lean
zrlk: Left a comment; it's thanks to Meta.evalExpr.

Kraken/X64/Tactics.lean
zrlk: Done.

Kraken/X64/Tactics.lean
zrlk: Done.

protz:
Should I take another look at this?

zrlk:
I was pretty happy with it after addressing the comments, but I figured I would run it through the Claude Opus model (which seems to do a much better job with Lean)... and it did a much better job. I think this is ready for review.

protz (commented):
Looks pretty good.
- Why do we need unsafe? The comment says evalExpr is unsafe, but it doesn't why it's unsafe and/or why we need it. (It would be of course preferable to not use unsafe, but I'm sure there's a good reason to do so.)
- delabListSpine -- is there nothing like it in the lean standard library already? I would assume this is a common helper

other than that looks good! let's investigate these two tidbits then merge, right now, I'm wishing I had this pr really badly :-)

zrlk:
Concerning `unsafe`: there's more detail in the comment, but the toplevel reasons are that without evalExpr, we would need to lift the rest of the printing infrastructure to the Expr level; and (besides divergence &c), evalExpr is unsafe because it performs an `unsafeCast` (we trust what we get is the representation of a `Directive`).

There does not appear to be anything quite like delabListSpine in the standard library. By rolling our own here we're able to (among other things) preserve location information in the output and print sequences of directives like `[asm| ...] ++ rest` when `rest` is unknown.

protz (approved):

protz:
Thanks, that's going to make everyone's life hella easier.

protz:
Does this also apply to the arm64 backend? If not, we should probably file a followup issue

zrlk:
Filed #162.

PR #137, opened 2026-08-06
Unified Stateful Semantics + Compiled Example

SamuelSchlesinger:
Exploring a more unified semantics and a compiled example that works uniformly across the architectures.

Marked draft for now, I am not looking to merge this as is. Let me know if a portion of it is wanted, but I wanted to post it for visibility.

### DIFF
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -20,6 +20,8 @@ jobs:
       - uses: actions/checkout@v5
       - uses: leanprover/lean-action@v1
       - run: python3 Kraken/X64/Test/asm_tests.py Kraken/X64/Test/asm
+      - name: Verify compiled sum-to-n artifact
+        run: python3 Kraken/Examples/check_sum_to_n.py --arch x64
 
   run-aarch64-asm-tests:
     name: Run AArch64 ASM tests (Ubuntu)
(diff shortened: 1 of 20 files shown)


protz (commented):
Some general thoughts about this PR.
- We initially went the route of `dsimp`, `simp` and other regular tactics. This is in the git history. The main issue with this is performance, along with controlling reduction finely. This is why we are writing the `kstep` tactic (see branch `protz/tactics` for the latest version) in the SymM monad, to get better perf for symbolic execution.

https://leodemoura.github.io/static/svil26/ has some slides about SymM, how it boosts performance, and how it is a suitable monad for symbolic execution.

The main blocking thing (as you saw, I believe) is that we do not currently have a way to advance just a few steps at a time. There is an attempt in #140 -- the most useful task at this stage would be to resolve #140, use this to prove `p3` from the examples with `kstep`, and use that as a baseline for conducting proofs within SymM, while getting the performance benefits.

If you have cycles to help out, that'd be the next item on the docket!

- How much of this is LLM-generated? We have an informal policy that should probably be made more explicit -- I can work on this, but it'd be great if you could clarify whether the design is yours or whether the design is auto-generated.

Thanks!

SamuelSchlesinger:
> There is an attempt in https://github.com/AeneasVerif/kraken/pull/140 -- the most useful task at this stage would be to resolve https://github.com/AeneasVerif/kraken/pull/140, use this to prove p3 from the examples with kstep, and use that as a baseline for conducting proofs within SymM, while getting the performance benefits.

I can take a look at this.

This code was all generated by AI, though I reviewed and steered some bits. Happy to make that more transparent going forward.

PR #138, opened 2026-08-06
fix(x64): properly parse + print displacements for rip-relative mem operands

zrlk:
We were correctly handling this in parseRelRegOrMem, but not in parseMemory.

Why didn't the roundtrip test catch this?

Consider this test program:

```asm
  lea (%rip), %rax
next:
  lea next(%rip), %rcx
  subq %rcx, %rax
  movl $0, %ecx
```

Prior to the parser fix, this is the Intel syntax we emit:

```asm
.intel_syntax noprefix
lea rax, [rip+0]
next:
lea rcx, [rip+next]
sub rax, rcx
mov ecx, 0
```

Unfortunately, `[rip+next]` happens to be the GNU assembler version of `next(%rip)`, so we generate an equivalent program despite not having built the proper representation. (side note: I'm pretty certain this logic does belong in the parser, not in Semantics.)

After the parser fix, we instead generate:

```asm
.intel_syntax noprefix
lea rax, [rip+0]
next:
lea rcx, [rip+(next - .)]
sub rax, rcx
mov ecx, 0
```

Now it doesn't roundtrip! The second `lea` changes from:

`-0x7(%rip),%rcx`

to

`0x0(%rip),%rcx`

It turns out that we also need to make a symmetrical fix in PrintIntel, where we also special-case `.rel (.sub e .after_current_instruction)`.

### DIFF
--- a/Kraken/X64/Parser.lean
+++ b/Kraken/X64/Parser.lean
@@ -367,6 +367,12 @@ def parseMemory : Parser (Width × AddrExpr) := do
     | w1, .none =>
       .pure w1
   let idx := Option.map (fun (_, idx) => ⟨idx, scale⟩) idx
+
+  -- Handle rip-relative addressing (like parseRelRegOrMem below).
+  let disp := match base, disp with
+    | .rip, .label l => .sub (.label l) .after_current_instruction
+    | _, _ => disp
+
   pure (w, { base, idx, disp })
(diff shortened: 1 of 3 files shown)


protz (approved):

protz:
Nice catch!

PR #139, opened 2026-08-06
Implement ecancel with shallow reflection

joehattori:
Reify separating conjunctions into shallow predicate trees, cancel
matching predicates, and instantiate residual metavariables. Construct
permutation proofs that certify the resulting equality.

### DIFF
--- a/Kraken/Separation.lean
+++ b/Kraken/Separation.lean
@@ -10,6 +10,8 @@ def sep (p q : ExtHashMap key value → Prop) (m : ExtHashMap key value) : Prop 
 infixl:60 " ⋆ " => sep
 notation:70 m " =⋆ " P => P m
 
+def emp : ExtHashMap key value → Prop := fun m => m = ∅
+
 omit [LawfulBEq key] in
 theorem disjoint_symm {m1 m2 : ExtHashMap key value} (h : m1.inter m2 = ∅) :
     m2.inter m1 = ∅ := by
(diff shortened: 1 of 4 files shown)


protz:
This seems great -- I'll take a look, and let's plan on getting a review from Andres once he comes back from vacation.

andres-erbsen:
FWIW I don't think you need to wait for me. Having this merged (after review by Jonathan) seems more valuable than having it perfect. (Maybe compare SymM usage against [sgraf's fresh canceller](https://github.com/leanprover/lean4/blob/42a1d764b58761bdda1a175625d4e9dd8e998d9a/tests/elab/vcgenSepLogic.lean#L632-L800) for inspiration while you're at it.(

protz:
Thanks Joe this is great. Let's wait until we merge #144, then here's what I suggest.

1. We make sure this PR is under regression testing. Let's pick one or two places where `ecancel` might be useful, for instance here: https://github.com/AeneasVerif/kraken/blob/93f33808eccf45e771e7c874a5452a5263ffb99d/Kraken/X64/Examples/Examples.lean#L359 We currently call `rw [sep_assoc] at h_mem` at line 347. Can we remove this call and call `ecancel` instead?
2. Then, merge this PR, knowing that it is under CI.
3. After #146 is fixed, we can try having `kstep` directly call `ecancel` as a discharger for memory-related obligations, without going through `tactic => ecancel`.

protz:
Question: can you explain where the backtracking monad is used? I'm concerned that Meta.saveState is going to be an expensive operation.

joehattori:
Thanks for the review. I updated the commits. 

> We make sure this PR is under regression testing. Let's pick one or two places where ecancel might be useful, for instance here:

The line in Examples.lean now uses `ecancel`. 

> Question: can you explain where the backtracking monad is used? I'm concerned that Meta.saveState is going to be an expensive operation.

The backtracking support is used in two places:

- `cancelClauses`: Matching may assign metavariables, and if cancelation of the remaining clauses fails, those assignments are undone and the DFS tries another pairing.
- `proveSeqEq`: `commitWhenSomeNoEx?` undoes changes made during proof construction upon failure.

IIUC, `Meta.saveState` doesn't perform deep copy so it should be relatively cheap. It is still not free, but the current Examples.lean file builds in about 4s (around the same watermark without ecancel), so I think the cost is reasonable.

protz (approved):

protz:
This is great. Thanks!

PR #140, opened 2026-08-06
WIP -- trying to relate step1 to straightLine

protz:
I am trying to relate step1 to straightline, so as to be able to step just a few instructions at a time -- namely, those instructions that are found at a given address.

Right now, this prevents us from stepping through enough instructions that we can do some loop-based reasoning. In short, restoring `p3` from the Examples directory.

This would also be useful for #137 which, as far as I can tell, does not use `kstep`.

I did some digging and found some issues with the layout -- these were fixed in my previous PR #136. I also rephrased directivesAtAddress and directivesFromAddress to make it obvious that the former is a prefix of the latter.

Sketching the proof, it turns out we need a monotonicity lemma for Directives.interp -- this is courtesy of Gemini.

The final step is still missing, hence why this PR is still draft.

### DIFF
--- a/Kraken/X64/OmniSemantics.lean
+++ b/Kraken/X64/OmniSemantics.lean
@@ -427,6 +427,17 @@ theorem Directives.interp_mono [Labels]
       hret
       h
 
+theorem Directives.interp_step_all [Labels]
+    (ds : List (Directive × Nat)) (s : MachineData) (pc : Int64)
+    (cont : Int64 → MachineData → Effects)
+    (post₁ post₂ : MachineState → Prop)
+    (hcont : ∀ pc' s', post₁ (s', pc') → (cont pc' s').All post₂)
+    (h : (Directives.interp ds s pc (fun pc' s' => Effects.done (s', pc'))).All post₁) :
+    (Directives.interp ds s pc cont).All post₂ := by
+  apply Directives.interp_mono ds s pc (fun pc' s' (h_ret : (Effects.done (s', pc')).All post₁) => by
+    dsimp [Effects.All] at h_ret
+    exact hcont pc' s' h_ret) h
+
 
 -- NOTE: 'initial' cannot be moved to the left of the colon as a parameter
 -- because it varies in the recursive call in the 'step' constructor (it becomes 'mid').
@@ -495,6 +506,44 @@ theorem directivesAtFromPrefix (e: Executable) (a: Int64):
   rw [← List.map_append]
   rw [List.takeWhile_append_dropWhile]
 
-theorem eventually_step [Layout] (e: Executable) (s: MachineState) (post: @Post MachineState):
-    step1 e s (fun s => straightlineStep e s post) → straightlineStep e s post := by
-  sorry
+theorem Directives.interp_append [Labels]
+    (ds1 ds2 : List (Directive × Nat)) (s : MachineData) (pc : Int64)
+    (ret : Int64 → MachineData → Effects)
+    {post : MachineState → Prop}
+    (hjmp : ∀ pc' s', (Directives.interp ds2 s' pc' ret).All post → (ret pc' s').All post)
+    (h : (Directives.interp ds1 s pc (fun pc' s' => Directives.interp ds2 s' pc' ret)).All post) :
+    (Directives.interp (ds1 ++ ds2) s pc ret).All post := by
+  induction ds1 generalizing s pc with
+  | nil =>
+    dsimp [Directives.interp] at *
+    exact h
+  | cons head tail ih =>
+    obtain ⟨d, sz⟩ := head
+    dsimp [Directives.interp] at *
+    apply Directive.interp_mono (jmp₁ := fun pc' s' => Directives.interp ds2 s' pc' ret) (jmp₂ := ret)
+      d s (.mk pc (pc + .ofNat sz))
+      (fun s' => ih s' (pc + .ofNat sz))
+      hjmp
+      h
+
+theorem eventually_step [Layout] (e: Executable) (st: MachineState) (post: @Post MachineState):
+    step1 e st (fun s => straightlineStep e s post) → straightlineStep e st post := by
+  intro h
+  let _ : Labels := e.labels
+  obtain ⟨s, pc⟩ := st
+  dsimp [step1, straightlineStep, Executable.step, Executable.straightline] at *
+  have hprefix := directivesAtFromPrefix e pc
+  rw [hprefix]
+  have h_step := Directives.interp_step_all (e.directivesAtAddress pc) s pc
+    (cont := fun pc' s' => Directives.interp (e.directivesFromAddress pc') s' pc' (fun pc s => Effects.done (s, pc)))
+    (post₁ := fun s' => (Directives.interp (e.directivesFromAddress s'.2) s'.1 s'.2 (fun pc s => Effects.done (s, pc))).All post)
+    (post₂ := post)
+    (hcont := fun pc' s' h => h)
+    h
+  apply Directives.interp_append (e.directivesAtAddress pc) _ s pc (fun pc s => Effects.done (s, pc))
+  · intro pc' s' h_jmp
+    dsimp [Effects.All]
+    sorry
+  · apply Directives.interp_mono (e.directivesAtAddress pc) s pc _ h_step
+    intro pc' s' h_cont
+    sorry
--- a/Kraken/X64/OmniSemantics.lean
+++ b/Kraken/X64/OmniSemantics.lean
@@ -90,3 +90,7 @@ theorem directivesAtFromPrefix (e: Executable) (a: Int64):
   refine ⟨((e.withAddresses.dropWhile (·.1 ≠ a)).dropWhile (·.1 = a)).map (·.2), ?_⟩
   rw [← List.map_append]
   rw [List.takeWhile_append_dropWhile]
+
+theorem eventually_step [Layout] (e: Executable) (s: MachineState) (post: @Post MachineState):
+    step1 e s (fun s => straightlineStep e s post) → straightlineStep e s post := by
+  sorry
--- a/Kraken/X64/OmniSemantics.lean
+++ b/Kraken/X64/OmniSemantics.lean
@@ -24,6 +24,410 @@ def Effects.All (post : MachineState → Prop) : Effects → Prop
   | .require_write_access _ _ cont => (cont ()).All post
   | .require_exec_access _ cont => (cont ()).All post
 
+theorem MachineData.load_mono {s : MachineData} {addr : BitVec 64} {w : Width}
+    {ret₁ ret₂ : w.type → MachineData → Effects} {post₁ post₂ : MachineState → Prop}
+    (hret : ∀ v s', (ret₁ v s').All post₁ → (ret₂ v s').All post₂)
+    (h : (s.load addr w ret₁).All post₁) :
+    (s.load addr w ret₂).All post₂ := by
+  dsimp [MachineData.load, Effects.All] at *
+  split at h
+  · exact hret _ _ h
+  · contradiction
+
+theorem MachineData.loadAvx_mono {s : MachineData} {addr : BitVec 64} {w : AvxWidth}
+    {ret₁ ret₂ : w.type → MachineData → Effects} {post₁ post₂ : MachineState → Prop}
+    (hret : ∀ v s', (ret₁ v s').All post₁ → (ret₂ v s').All post₂)
+    (h : (s.loadAvx addr w ret₁).All post₁) :
+    (s.loadAvx addr w ret₂).All post₂ := by
+  dsimp [MachineData.loadAvx, Effects.All] at *
+  split at h
+  · exact hret _ _ h
+  · contradiction
+
+theorem MachineData.store_mono {s : MachineData} {addr : BitVec 64} {w : Width} {v : w.type}
+    {ret₁ ret₂ : MachineData → Effects} {post₁ post₂ : MachineState → Prop}
+    (hret : ∀ s', (ret₁ s').All post₁ → (ret₂ s').All post₂)
+    (h : (s.store addr v ret₁).All post₁) :
+    (s.store addr v ret₂).All post₂ := by
+  dsimp [MachineData.store, Effects.All] at *
+  split at h
+  · exact hret _ h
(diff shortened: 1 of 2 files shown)


Kraken/X64/Examples/Examples.lean
```diff
@@ -20,12 +20,40 @@ import Kraken.X64.Sep
...
+  Nat.shiftRight_zero
+  Nat.sub_zero
+  UInt64.ofBitVec_add
+  UInt64.ofBitVec_ofNat
+  UInt64.ofBitVec_sub
+  UInt64.ofBitVec_toBitVec
+  UInt64.sub_add_cancel
+  UInt64.toBitVec_ofNat
+  UInt64.toBitVec_sub
+  UInt64.toNat_toBitVec
```
andres-erbsen, line 47: Could we use attribute `grind_hom` instead of generic simp to handle these lemmas?

Kraken/X64/Examples/Examples.lean
```diff
@@ -20,12 +20,40 @@ import Kraken.X64.Sep
...
+
 --------------------------------------------------------------------------------
 
 def p1 := parse("start: mov $1, %rax")
 
 -- Super-simple example to debug tactics
-example [layout : Layout] s : straightlineStep (layout p1) (s, layout.start) (fun s => s.1.regs.rax = 1) := by
+example [layout : Layout] (hwf : (layout p1).WellFormed) s :
```
andres-erbsen, line 54: > (hwf : (layout p1).WellFormed)

We should be able set up the semantics so that:
- a program with instructions larger than 16 bytes admits no steps (conservatively treating oversized-instruction execution as undefined behavior)
- a program with less than 2^64/16 instructions is automatically proven well-formed

Kraken/X64/OmniSemantics.lean
```diff
@@ -17,8 +17,547 @@ import Kraken.X64.Semantics
...
+        simp [hx_pc]
+      omega
+    exact eventually_step e hwf (s', y.1) post h_straightline_next
+termination_by (e.withAddresses.dropWhile (·.1 ≠ st.2)).length
+decreasing_by exact h_len
+
+theorem eventually_step_cps [Layout] (e : Executable) (hwf : e.WellFormed)
+    (st : MachineState) (post : @Post MachineState) :
+    straightlineStep e st (fun mid => Eventually (step1 e) post mid) →
+    Eventually (step1 e) post st := by
```
andres-erbsen, line 536: The name of this lemma does not match the abstraction level of the statement. Both generic Omnisemantics.facts and X64-specific specializations can be useful, but we should keep the difference clear.

Leaving some arguments implicit/inferred in applications of this lemma could also improve readability of proof scripts

Kraken/X64/OmniSemantics.lean
```diff
@@ -17,8 +17,547 @@ import Kraken.X64.Semantics
...
+    obtain ⟨d, sz⟩ := head
+    dsimp [Directives.interp] at *
+    exact Directive.interp_mono d s (.mk pc (pc + .ofNat sz))
+      (fun s' => ih s' (pc + .ofNat sz) hnext)
+      hjmp
+      h
+
+theorem eventually_step [Layout] (e: Executable) (hwf : e.WellFormed) (st: MachineState) (post: @Post MachineState):
+    straightlineStep e st post →
+    Eventually (step1 e) post st
```
andres-erbsen, line 479: Rename to `eventually_straightlineStep`? The current name reads like the premise is a single step.

Kraken/X64/OmniSemantics.lean
```diff
@@ -17,8 +17,547 @@ import Kraken.X64.Semantics
   | .require_write_access _ _ cont => (cont ()).All post
   | .require_exec_access _ cont => (cont ()).All post
 
+theorem MachineData.load_mono {s : MachineData} {addr : BitVec 64} {w : Width}
```
andres-erbsen, line 20: All the mechanical target-specific semantics-well-formedness lemmas should probably go into a dedicated file

andres-erbsen (commented):
Partial review

Kraken/X64/Examples/Examples.lean
protz: decision: use `Fin 16` for `instr_size` and then derive well-formedness automatically (by `decide` or something)

Kraken/X64/OmniSemantics.lean
protz: todo: hoist this to a generic lemma that is over `{MachineState}` -- generally, do we need two OmniSemantics files? pinging @lukaszobernig because I think this is from your earlier refactoring PR, I forgot why we ended up this way

Kraken/X64/OmniSemantics.lean
protz: decision: categorize lemmas into
1. completely mechanical stuff that could be written by an AI or a meta-program (monotonicity lemmas)
2. accounting / book-keeping lemmas e.g. for the behavior of withAddresses, reverse and fold
3. high-level meaningful lemmas for stepping, to be used by high-level tactics

Category 3 can be further split between machine-specific and generic lemmas.

We can also use the upcoming switch to Lean's module system to make 1. and 2. private (insofar as possible). To be investigated.

Kraken/X64/Examples/Examples.lean
```diff
@@ -95,108 +124,135 @@ _end:
...
 
-set_option maxHeartbeats 4000000 in
-theorem p3_correct [layout: Layout] (s: MachineData):
-    p3_spec s < 2^64 →
-    Eventually (straightlineStep (layout p3)) (fun s => s.1.regs.rdx.toNat = p3_spec s.1 ∧ s.1.regs.rax = 0) (s, layout.start) :=
-  by
-    intros h_bounds
-    apply step_cps
-    kprologue p3 with s
+private theorem int64_rel_jmp_target (u tgt : Int64) :
```
protz, line 127: TODO: investigate why this (and other helpers below) are not dischargeable through `grind`

Kraken/X64/Examples/Examples.lean
```diff
@@ -95,108 +124,135 @@ _end:
...
+  kstep 2
+  tactic =>
+  change Effects.All _ (@Directives.interp (Executable.labels (layout p3)) ((p3.mapIdx (fun i d => (d, layout.size i))).drop 2) _ pc_start (fun pc s => Effects.done (s, pc)))
+  rw [← h_from_start]
+  change straightlineStep (layout p3) (_, pc_start) _
+  dsimp only [p3_spec] at hb ⊢
+  apply tailrec_loop_straightline (layout p3) h
+    (fun s' => s'.1.regs.rdx.toNat = 2 ^ (2 ^ rbx.toNat) ∧ s'.1.regs.rax = 0)
+    (_, pc_start)
+    (fun v st =>
```
protz, line 186: do we need to state this extensive post-condition or can we rephrase the lemma in a way that the postcondition (at least for trivial cases) can just be `_`?

Kraken/X64/Examples/Examples.lean
```diff
@@ -95,108 +124,135 @@ _end:
...
-                rw [h_vi_k, Nat.mod_eq_of_lt]
-                . rw [← Nat.pow_two, ← Nat.pow_mul, ← Nat.pow_succ]
-                . apply Nat.lt_of_le_of_lt _ h_bounds
-                  rw [← Nat.pow_two, ← Nat.pow_mul, ← Nat.pow_succ]
-                  apply Nat.pow_le_pow_right (by decide)
-                  apply Nat.pow_le_pow_right (by decide)
-                  omega -/
+    rfl
+  · intro v state ⟨hpc, hrbx, hle, hrdx, hrax_st⟩
+    obtain ⟨⟨⟨rax', rbx', rcx', rdx', rsi', rdi', rsp', rbp', r8', r9', r10', r11', r12', r13', r14', r15'⟩, zmms', flags', mem'⟩, pc'⟩ := state
```
protz, line 197: again? is this necessary in the proof script?

Kraken/X64/Examples/Examples.lean
```diff
@@ -95,108 +124,135 @@ _end:
...
+    dsimp only at hpc hrbx hrdx hrax_st
+    subst hpc
+    dsimp only [straightlineStep, Executable.straightline]
+    rw [h_from_start]
+    delta p3
```
protz, line 202: why is this necessary? investigate

Kraken/X64/Examples/Examples.lean
```diff
@@ -95,108 +124,135 @@ _end:
...
+    rename_i v_sub status
+    have hv_lt : v < 2 ^ 64 := hrbx ▸ rbx'.toBitVec.isLt
+    have hv_sub : v_sub = BitVec.ofNat 64 v := by simp [v_sub, ← hrbx]
+    by_cases hv0 : v = 0
+    · have h_cond : (v_sub == BitVec.zero 64) = true := by rw [hv_sub, hv0]; rfl
+      simp only [h_cond, ↓reduceIte, Effects.All]
+      exact Or.inl ⟨by rw [hrdx, hv0, Nat.sub_zero], hrax_st⟩
+    · have h_cond : (v_sub == BitVec.zero 64) = false := by
+        rw [hv_sub]
+        apply Bool.eq_false_iff.mpr
```
protz, line 216: TODO: investigate whether with prompting that relies on the PR for grind homomorphisms and documentation for kstep, we can make this proof more concise

Kraken/X64/OmniSemantics.lean
```diff
@@ -17,8 +17,547 @@ import Kraken.X64.Semantics
...
+      hret
+      h
+
 def step1 [Layout] (e: Executable) (s: MachineState) (post: @Post MachineState) : Prop :=
   (Executable.step e s .done).All post
 
 def straightlineStep [Layout] (e: Executable) (s: MachineState) (post: @Post MachineState) : Prop :=
   (Executable.straightline e s .done).All post
+
+theorem Directives.interp_split [Labels]
```
protz, line 455: mark as private -- doesn't make much sense to me, I was expecting the other direction?

Kraken/Layout.lean
```diff
@@ -64,10 +81,242 @@ theorem Executable.directivesFromStart {Directive : Type} [layout : Layout Direc
...
+        rw [h_drop_a1, List.dropWhile_cons, List.dropWhile_cons]
+        simp [ha2_ne_a1']
+      exact hwf a1 (a2, d2, sz2) suf h_step2
+  dsimp [Executable.directivesFromAddress]
+  change (e.withAddresses.dropWhile (·.1 ≠ a2)).map (·.2) = (d2, sz2) :: rest
+  rw [h_drop_a2, List.map_cons]
+  congr 1
+  exact Executable.withAddresses_map_snd rest (a2 + Int64.ofNat sz2)
+
+private theorem int64_add_ofNat_assoc (a : Int64) (m n : Nat) :
```
protz, line 191: same remark: why not grind

Kraken/Layout.lean
```diff
@@ -64,10 +81,242 @@ theorem Executable.directivesFromStart {Directive : Type} [layout : Layout Direc
...
+      obtain ⟨hne, k, hk_le, hy⟩ := withAddresses_dropWhile_eq_offset start_addr start_addr ((d, sz) :: tl) h'
+      have hk_pos : 0 < k := by
+        cases k with
+        | zero =>
+          simp at hy
+          exact absurd hy hne
+        | succ n => omega
+      exact ⟨k, hk_pos, hk_le, hy⟩
+
+theorem Executable.wellFormed_of_sum_lt {Directive : Type} (e : Executable Directive)
```
protz, line 262: this is the central lemma and we should add a comment

protz (commented):
To fix before merging:
- investigate a little bit more the large amount of lemmas related to withAddresses, to separate between stuff that is directly in service of obtaining the well-formedness lemma from instruction bounds, and stuff that is equational theory (can it be improved?)
- time-bounded attempt to see if we can leverage grind homomorphisms

The rest can be addressed with followups

Kraken/X64/OmniSemantics.lean
lukaszobernig: I remember trying to make this generic but it ended up being more code. And since Effects are different, each architecture got its own minimal OmniSemantics file to handle just that. But I suppose we can eat the extra lines now and make it generic.

andres-erbsen: IMO Omnisemantics should also be generic over effects, even if it makes for more lines. 

Is there a deep reason why effects are different? Conceptually, if we had heterogeneous multi-core system with ARM64 and X64 cores, would they genuinely have different bus-level capabilities?

lukaszobernig: Here x64 has a `gp_unaligned` (for unaligned vector loads/stores), AArch64 has a `unaligned_sp` (for unaligned SP during LDP/STP). We could unify these into an `alignment_check` effect. Or have a more generic "error interrupt" mechanism to capture such failures.

andres-erbsen: Both of these sound like "ErrorUnimplemented" stand-ins for the real CPU-local behavior, without deep conceptual differences. I'd lean towards sharing Effects.

protz: LLM generalization based on type classes in protz/gen-mono

Kraken/X64/OmniSemantics.lean
```diff
@@ -17,8 +17,547 @@ import Kraken.X64.Semantics
...
+    {p q : @Post MachineState} (hpq : ∀ s, p s → q s) :
+    straightlineStep e st p → straightlineStep e st q :=
+  let _ := e.labels
+  Directives.interp_mono _ _ _ (fun _ _ => hpq _)
+
+theorem tailrec_loop_straightline [Layout] (e : Executable) (hwf : e.WellFormed)
+    (post : @Post MachineState) (initial : MachineState)
+    (P : Nat → @Post MachineState) (v0 : Nat) (hP : P v0 initial)
+    (hbody : ∀ v state, P v state →
+      straightlineStep e state (fun mid_s => post mid_s ∨ ∃ v', P v' mid_s ∧ v' < v)) :
```
andres-erbsen, line 552: This doesn't apply when there's even an `if` inside the loop, does it? (I'm not sure we want a lemma that limited, but not strongly opposed either)

PR #141, opened 2026-08-10
fix: work around an issue with kdsimpProj / Sym.DSimp.dsimpProj

zrlk:
Machine-generated reduction/fix; the accusation against upstream is the robot's, not mine.


zrlk:
This was actually fixed in 4.33.0, so with #144 we don't need it.

PR #144, opened 2026-08-10
Many improvements in the `kstep` tactic

protz:
It took a while before I could restore examples and I didn't want to merge something that regressed, so this is why this PR is a little big. (Still need to merge main into this branch)

Highlights:
- improved kstep tactics that now iterates three phases of
  * dsimp -- reducing the semantics over the program in the goal, reducing all definitions marked as `@[kstep]`
  * simp -- applying relevant theorems to simplify the goal, via a new attribute `@[ksimp]`
  * spec -- finding lemmas tagged with `@[kspec]`, and running `rw` over them to advance the goal via a specification step (typically: `load_sep` or `store_sep`)
- a corollary is that `kstep` now generates subgoals; the strategy for subgoals is now to
  * identify the subgoal that is a separation logic subgoal and try to solve it (this is where @joehattori's ecancel will come in -- right now this is a dumb search for a hypothesis that `exact`ly matches)
  * if the separation logic subgoal is solved, great, solve the rest of the subgoals (using a combination of rfl, exact, or grind)
  * if the separation logic subgoal is not solved, abort and return control to the user, rather than make dumb decisions in other subgoals (e.g. instantiating metavariables with any hypothesis)
- `kstep` is now also capable of `intro`ducing let-binders in the goal -- it would be easier if we had a Sym `lift_lets` but that'll do for now
- `kstep` takes a maximum number of directives to step through; this is useful, but will be even more useful in conjunction wtih #140 

I need to do some minor cleanup, merge main into this, and then this PR should hopefully be good and pave the way for merging @joehattori's `ecancel`. 

There are a myriad of small-ish issues that make this less than usable. I will be filing issues for all of those shortly.


Kraken/X64/Examples/Examples.lean
```diff
@@ -432,9 +380,11 @@ def sib_example := parse("
     movq (%rdi, %r15, 8), %rax
 ")
 
+-- FIXME: I had to replace `s₀.regs.r15.toBitVec * 8#64` with `BitVec.ofInt 64
+-- (s₀.regs.r15.toBitVec.toInt * 8)` to make the exampel go through. Why?
```
zrlk: nit: example

zrlk (approved):
This should help a lot, thanks!

PR #150, opened 2026-08-11
Error out if kstep cannot step through the requested number of instructions

protz:
This is better usability than silently failing to advance


zrlk (approved):
Thanks! I got caught on this a couple of times. Getting explicit feedback that something went wrong will be helpful.

PR #151, opened 2026-08-11
Add named registers to kprologue

### LINKED ISSUE #145: Improve kprologue to name registers
So I remembered why I struggled with `kprologue` -- it is because I wanted the macro to also do this:

```lean4
  let ss := s₀
  change (straightlineStep _ (ss, _) _)
  cases s₀ with | mk regs zmms flags mem =>
  cases regs with | mk rax rbx rcx rdx rsi rdi rsp rbp r8 r9 r10 r11 r12 r13 r14 r15 =>
```

which is basically a trick to give pretty names to all the registers and have them in the context, rather than having to write expressions like s.regs.rax and so on.

I think this is generally useful and something we should do for every program proof. But the `cases ... => ...`, I remembered, posed difficulties in terms of how to hoist these four lines into a macro while still keep hygiene.

SamuelSchlesinger:
Closes #145.

I figured it is nicer if this is robust to changes in the underlying machine architecture, so I did it in this way. There is a simpler way of course, where we enumerate the registers by hand, but it could introduce annoying migration issues down the line if it isn't updated appropriately.

### DIFF
--- a/Kraken/X64/Examples/Examples.lean
+++ b/Kraken/X64/Examples/Examples.lean
@@ -23,7 +23,7 @@ def p1 := parse("start: mov $1, %rax")
 
 -- Super-simple example to debug tactics
 example [layout : Layout] s : straightlineStep (layout p1) (s, layout.start) (fun s => s.1.regs.rax = 1) := by
-  kprologue p1
+  kprologue p1 with s
   sym => kstep; tactic =>
   decide
   /- simp [Instr.interp,Operation.interp,Operand.interp,MachineData.set] -/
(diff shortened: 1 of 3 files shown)


protz (changes requested):
This is great. The only change I would request is maybe to remove step_cps from the tactic, or at least make it something we can disable.

Once #140 is fixed we might want to do things other than step_cps, to break the `eventually` into two 

Thanks!!

SamuelSchlesinger:
> Once https://github.com/AeneasVerif/kraken/pull/140 is fixed we might want to do things other than step_cps, to break the eventually into two

Indeed, we do some different things in my local branch. I'll push a PR up for that soon.

Also, I split out `step_cps`, should be good to merge now if you'd like.

protz (approved):

PR #154, opened 2026-08-13
Factor out common tactics and lemmas for X64 and AArch64

lukaszobernig:
This also adds some examples to AArch64 that make use of the tactics machinery.

I have also taken the liberty to clean up the include order and the files I touched. I will clean up the header comments a bit more in a follow up.


Kraken/OmniSemantics.lean
```diff
@@ -0,0 +1,51 @@
...
+    | step initial mid_p ht _ ih => exact .step initial mid_p ht ih
+
+theorem eventually_weaken {State : Type} (trans : State → Post → Prop) (p q : Post) (initial : State)
+  (h : ∀ s, p s → q s) :
+    Eventually trans p initial → Eventually trans q initial
+  := by
+    exact fun hp => eventually_trans trans p q initial hp fun s hs => .done s (h s hs)
+
+-- A loop down to 0
+theorem reg_dec_loop {State : Type} (trans : State → Post → Prop) (post : Post) (initial : State) (invariant : Nat → Post) (n : Nat) :
```
miriampolzer, line 41: nit: the previous reg_dec_loop had some comments, should we put them here as well?

Kraken/Tactics.lean
```diff
@@ -1,46 +1,36 @@
...
+import Lean
+import Std
+
+open Lean Meta Elab Tactic Kraken
 
 /-- Set up a straight-line proof for `p` from `s`, naming the state fields and
 general-purpose registers. `status` and `dmem` are called `flags` and `mem`. -/
-syntax (name := kprologue) "kprologue" ident "with" ident : tactic
+syntax (name := kprologueWith) "kprologue" ident "with" ident : tactic
+syntax (name := kprologueSimple) "kprologue" ident : tactic
```
miriampolzer: I suppose this is a version of kprologue that doesn't unfold a program? As far as I see it is not used yet. For which kind of example would we use this, and could you add one, also to avoid regressions?

miriampolzer (commented):
Looks good to me! But ofc let's wait for Jonathan's review as well, I am not proficient enough in kraken at the moment to parse all of this, especially the change you had to make to kstep.

Kraken/Tactics.lean
lukaszobernig: Oh no this is just me copying the wrong file 🙈  thanks for catching!

Kraken/OmniSemantics.lean
lukaszobernig: Good point, I've added them back!

Kraken/Tactics.lean
lukaszobernig: Fixed, thank you!

Kraken/Tactics.lean
```diff
@@ -417,9 +398,19 @@ partial def evalSymKStep : Grind.GrindTactic :=
     let goalState ← do
       let goalT ← goal.mvarId.getType
       let_expr gimmickId goalT' := goalT | throwError "missing gimmick"
+      let goalT' ← instantiateMVars goalT'
```
protz, line 401: why do you need to instantiate mvars here?

protz (approved):
This is great. If you could elucidate the call to instantiateMVars before merging, that'd be great. Just a comment will do. Thank you!

Kraken/Tactics.lean
lukaszobernig: After the arch split, we don't explicitly have `Effects.All` anymore (before it was imported through the x64 OmniSemantics.lean file). And since mvar assignments are lazily stored we need to instantiate here before it's even possible to syntactically match for `Effects.All` in all cases. Otherwise we would be falling through a bunch of `Expr.mvar ?m` where `?m` could have been assigned as `?m := Effects.All ...` previously somewhere. I hope that makes sense.

protz: it does, thank you -- might be worth adding as a comment in the source, but I'll merge now so that I don't hold up your work

Issue #18, opened 2026-03-05
Fix Broken Proof of Example p3


aferr:
This proof which we previously hard working is now a sorry: https://github.com/AeneasVerif/kraken/blob/bc619dc8191968caca327e0ca97487cc1ef1d4f4/Kraken/Examples.lean#L72

The proof broke as of https://github.com/AeneasVerif/kraken/commit/40325303e79601736d3f911052ca7bfe44a157e3

The likely cause is the larger number of cases with register aliasing. It might be fixed by implementing #10

PR #156, opened 2026-08-17
Return control flow as data: relate step1 to straightline, prove p3

SamuelSchlesinger:
Long PR description incoming, but I felt being quite explicit was warranted given the size of this PR.

In #137, the suggested next step was to resolve #140 and use it to prove p3 from the examples with `kstep`. This does both, by changing the interpreters rather than relating the two CPS views after the fact: `Operation.interp` now produces `Effects (MachineData × Ctrl)`, `Directives.interp` folds a listing into `Effects (MachineData × BlockExit)`, and `Effects` is a lawful monad. `step` and `straightline` are continuation shells over that core, so `step1`, `straightlineStep`, `kprologue`, `eval`, and the runners keep their shapes, and every example statement and proof replays unchanged on both ISAs, except p3 (below).

With the exit as data, appending directives is an equation (`Directives.interp_append`) rather than the implication #140 got stuck on: its `hjmp` obligation asked a jump exit and an end-of-list exit to coincide, which fails whenever there's still work after the jump target. `straightline_eq_step` and the `Eventually`-level transfer from #104 follow. They carry a fetch-coherence side condition, and it's a substantive one since `Layout.size` permits aliasing (the bit #155 quarantines with `SaneP3Layout`); `Executable.Resolves` abstracts this and reduces to a finite check per program.

p3 is now proven. The old statement was false (its binder shadowed the initial state) and had been sorried since #18, so I adopted the corrected statement from @aferr's #155 and proved it with `kstep` and his address lemmas. Either landing order with #155 works.

The tactic layer is the same #144/#150 machinery but now it can walk binds: `kstep` proofs look exactly like they did before, the step budget gets its first regression test, and there's a new `kcut` that runs a bounded prefix and leaves a resume goal of the same shape.

Some evidence this is meaning-preserving: the old and new runners are bit-identical on all 34 x64 and 63 AArch64 test programs, every commit builds on its own, the new theorems use only the standard axioms, and elaboration times are roughly flat (the shared x64 examples pick up about 10%). One caveat: `step` fetches by the size table, which differs from address-equality fetch only on layouts whose sizes wrap mod 2^64. The AArch64 conversion mirrors x64 case by case; the spots that deserve SDM eyes are LDP/LDPSW load order, STP's pre-store reads, and BLR's target-before-X30 write.

Left for later: the dynamic-stack sorry (a curated ksimp set should get it), a general `Resolves` witness from a total-size bound, and deduplicating `Effects` across the ISAs. This code was written mostly by a variety of AI assistants, while I steered the design and reviewed everything in detail (as opposed to #137, where I was a bit over my skis). The final copy was finished with Claude Fable 5 after many iterations with GPT 5.6 Sol and Claude Fable 5.

Happy to make this smaller or do it more incrementally if desired, but I figured this was a good place to cut it and get feedback where it gives you an idea of how this impacts the design as a whole.

### DIFF
--- a/Kraken/AArch64/OmniSemantics.lean
+++ b/Kraken/AArch64/OmniSemantics.lean
@@ -155,8 +155,7 @@ def Executable.ResolvesFallthroughAt (e : Executable) (pc : Int64) : Prop :=
         (Kraken.Directives.takeBlock (e.directivesFromAddress pc)) pc)
 
 /-- A straight-line run executes the block at the current address, then
-continues only on fall-through. A jump ends the run at its target. This is the
-sound form of the step/straightline relation PR #140 was after. -/
+continues only on fall-through. A jump ends the run at its target. -/
 theorem Executable.straightline_eq_step {α : Type}
     (e : Executable) (st : MachineState)
     (h : e.ResolvesFallthroughAt st.2)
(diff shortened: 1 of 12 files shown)


protz:
Thanks for submitting this! I'm glad it's out so that we can take a look and start discussing.

I'm about to head out for a vacation and won't have time to review this in depth before I go. This is a large PR, so I suspect it is going to take a few rounds of discussion to hash out what is happening and what is the fundamental reason why the proof of equivalence between step1 and straightlinestep is difficult (either here or in #155).

Perhaps @andres-erbsen will want to take a look once he's back online, or @samuelgruetter might have thoughts too since he authored the effect system initially.

Thanks.

SamuelSchlesinger:
As I mentioned in the thread, I can definitely split this up for review.

SamuelSchlesinger:
Oh, I should clarify: the proof in #140 is difficult because the statement is false. I forget the counter example and I am taking a mental health day today so not by computer.

protz:
Have a good day off, Sam, and if you ever remember what the counterexample was, it might help us understand what's going on!

SamuelSchlesinger:
`0: jmp target 1: target: 1: nop` is the minimal counterexample. Basically, `eventually_step` (in #140) jumps to `1`, then its continuation runs `straightlineStep` there, executes the `nop`, and finishes at `2`. But `straightlineStep` from `0` stops at the taken jump and returns at `1`.

So the problem is that the old CPS interface gives jumps and fallthrough an indistinguishable continuation. The suffix should run only after fallthrough, i.e. a jump should be passed directly to post. That is what `BlockExit` is meant to fix in this PR.

PR #157, opened 2026-08-18
Refactor semantics to use BitVec instead of unbounded integers

lukaszobernig:
This PR refactors the semantics to use `BitVec` instead of unbounded integers (`Int`/`Nat`). This should make it easier to use `bv_decide` and also keeps the semantics in terms of "64-bit machine words". For multiplication and flags computation we also include some helper lemmas and specs to rewrite the `BitVec` forms into arithmetic integer-based forms. This should help with reasoning using tactics like `lia`.

### DIFF
--- a/Kraken/AArch64/Parser.lean
+++ b/Kraken/AArch64/Parser.lean
@@ -409,11 +409,11 @@ Validate an optional shift amount for register operands:
 - Must be in `[0, 31]` for 32-bit instructions (`.W32`).
 - Must be in `[0, 63]` for 64-bit instructions (`.W64`).
 -/
-def checkShiftAmount (w : RegWidth) (amt : Int64) : Except String Unit :=
+def checkShiftAmount (w : RegWidth) (amt : Int64) : Except String Nat :=
   let maxAmt := w.bits - 1
   if amt.toInt < 0 || amt.toInt > maxAmt then
     .error s!"shift amount {amt.toInt} out of range [0, {maxAmt}] for {w.bits}-bit instruction"
-  else .ok ()
+  else .ok amt.toInt.toNat
 
 /--
 Validate a signed immediate byte offset for load/store pair instructions (`LDP`/`STP`):
@@ -616,11 +616,11 @@ def checkCcmpImmediate (imm : Nat) : Except String Unit :=
     .error s!"ccmp/ccmn immediate {imm} out of range [0, 31]"
   else .ok ()
 
-/-- Validate NZCV condition flags immediate operand (`[0, 15]`). -/
-def checkNzcvFlags (nzcv : Nat) : Except String Unit :=
+/-- Validate the NZCV condition flags immediate operand (`[0, 15]`) and return it as a 4-bit flags field. -/
+def checkNzcvFlags (nzcv : Nat) : Except String (BitVec 4) :=
   if nzcv > 15 then
     .error s!"nzcv immediate {nzcv} out of range [0, 15]"
-  else .ok ()
+  else .ok (BitVec.ofNat 4 nzcv)
 
 /-- Validate a 12-bit unsigned arithmetic immediate (for ADD/SUB immediate operands). -/
 def checkArithmeticImmediate (imm : Int64) : Except String Unit :=
@@ -951,8 +951,8 @@ def parseShiftRegExpr (w : RegWidth) (allowRor : Bool := false) : Parser (ShiftR
       | _ => fail s!"unknown shift type: {shiftName}"
     skipHWs
     let amt ← parseInt64
-    liftExcept (checkShiftAmount w amt)
-    pure { reg := reg, amount := amt, shift := shiftType }
+    let amount ← liftExcept (checkShiftAmount w amt)
+    pure { reg := reg, amount := amount, shift := shiftType }
   else
     pure { reg := reg, amount := 0, shift := ShiftType.LSL }
 
@@ -1846,8 +1846,8 @@ or immediate second operand.
 - Validates that `#imm` is in range `[0, 31]` and `#nzcv` flags immediate is in range `[0, 15]`.
 -/
 def parseCondCompare
-    (mkReg : {w : RegWidth} → RegOrZr w → RegOrZr w → Nat → CondCode → Operation w)
-    (mkImm : {w : RegWidth} → RegOrZr w → Nat → Nat → CondCode → Operation w) : Parser Instr := do
+    (mkReg : {w : RegWidth} → RegOrZr w → RegOrZr w → BitVec 4 → CondCode → Operation w)
+    (mkImm : {w : RegWidth} → RegOrZr w → Nat → BitVec 4 → CondCode → Operation w) : Parser Instr := do
   let src1W ← parseRegOrZrW
   let w := src1W.w
   parseComma
@@ -1856,16 +1856,16 @@ def parseCondCompare
     let imm ← parseImmNat
     liftExcept (checkCcmpImmediate imm)
     parseComma
-    let nzcv ← parseImmNat
-    liftExcept (checkNzcvFlags nzcv)
+    let nzcvNat ← parseImmNat
+    let nzcv ← liftExcept (checkNzcvFlags nzcvNat)
     parseComma
     let cond ← parseCondArg
     pure ⟨w, mkImm src1W.reg imm nzcv cond⟩
   else
     let src2 ← parseRegOrZr w
     parseComma
-    let nzcv ← parseImmNat
-    liftExcept (checkNzcvFlags nzcv)
+    let nzcvNat ← parseImmNat
+    let nzcv ← liftExcept (checkNzcvFlags nzcvNat)
     parseComma
     let cond ← parseCondArg
     pure ⟨w, mkReg src1W.reg src2 nzcv cond⟩
--- a/Kraken/AArch64/Parser.lean
+++ b/Kraken/AArch64/Parser.lean
@@ -121,6 +121,17 @@ def parseImmNat : Parser Nat := do
     fail s!"expected non-negative immediate, got {i}"
   pure i.toNat
 
+/--
+Parse an unsigned immediate integer value fitting into a bitvector of width `w`.
+- Fails if the parsed integer is negative or exceeds `2^w - 1`.
+-/
+def parseBitVec (w : Nat) : Parser (BitVec w) := do
+  let n ← parseImmNat
+  let maxVal := 2^w - 1
+  if n > maxVal then
+    fail s!"immediate {n} out of range [0, {maxVal}]"
+  pure (BitVec.ofNat w n)
+
 /-- Parse a raw label name as a string identifier. -/
 def parseLabelRaw : Parser Label := parseName
 
@@ -409,11 +420,11 @@ Validate an optional shift amount for register operands:
 - Must be in `[0, 31]` for 32-bit instructions (`.W32`).
 - Must be in `[0, 63]` for 64-bit instructions (`.W64`).
 -/
-def checkShiftAmount (w : RegWidth) (amt : Int64) : Except String Nat :=
+def checkShiftAmount (w : RegWidth) (amt : Nat) : Except String Unit :=
   let maxAmt := w.bits - 1
-  if amt.toInt < 0 || amt.toInt > maxAmt then
-    .error s!"shift amount {amt.toInt} out of range [0, {maxAmt}] for {w.bits}-bit instruction"
-  else .ok amt.toInt.toNat
+  if amt > maxAmt then
+    .error s!"shift amount {amt} out of range [0, {maxAmt}] for {w.bits}-bit instruction"
+  else .ok ()
 
 /--
 Validate a signed immediate byte offset for load/store pair instructions (`LDP`/`STP`):
@@ -615,12 +626,6 @@ def checkCcmpImmediate (imm : Nat) : Except String Unit :=
   if imm > 31 then
     .error s!"ccmp/ccmn immediate {imm} out of range [0, 31]"
   else .ok ()
-
-/-- Validate the NZCV condition flags immediate operand (`[0, 15]`) and return it as a 4-bit flags field. -/
-def checkNzcvFlags (nzcv : Nat) : Except String (BitVec 4) :=
-  if nzcv > 15 then
-    .error s!"nzcv immediate {nzcv} out of range [0, 15]"
-  else .ok (BitVec.ofNat 4 nzcv)
 
 /-- Validate a 12-bit unsigned arithmetic immediate (for ADD/SUB immediate operands). -/
 def checkArithmeticImmediate (imm : Int64) : Except String Unit :=
@@ -950,9 +955,9 @@ def parseShiftRegExpr (w : RegWidth) (allowRor : Bool := false) : Parser (ShiftR
         else fail "arithmetic instructions do not support ROR shift"
       | _ => fail s!"unknown shift type: {shiftName}"
     skipHWs
-    let amt ← parseInt64
-    let amount ← liftExcept (checkShiftAmount w amt)
-    pure { reg := reg, amount := amount, shift := shiftType }
+    let amt ← parseImmNat
+    liftExcept (checkShiftAmount w amt)
+    pure { reg := reg, amount := amt, shift := shiftType }
   else
     pure { reg := reg, amount := 0, shift := ShiftType.LSL }
 
@@ -1856,16 +1861,14 @@ def parseCondCompare
     let imm ← parseImmNat
     liftExcept (checkCcmpImmediate imm)
     parseComma
-    let nzcvNat ← parseImmNat
-    let nzcv ← liftExcept (checkNzcvFlags nzcvNat)
+    let nzcv ← parseBitVec 4
     parseComma
     let cond ← parseCondArg
     pure ⟨w, mkImm src1W.reg imm nzcv cond⟩
   else
     let src2 ← parseRegOrZr w
     parseComma
-    let nzcvNat ← parseImmNat
-    let nzcv ← liftExcept (checkNzcvFlags nzcvNat)
+    let nzcv ← parseBitVec 4
     parseComma
     let cond ← parseCondArg
     pure ⟨w, mkReg src1W.reg src2 nzcv cond⟩
--- a/Kraken/AArch64/Semantics.lean
+++ b/Kraken/AArch64/Semantics.lean
@@ -215,7 +215,8 @@ export Labels (label)
     let lo := val &&& 0xFFF#64
     Int64.ofBitVec lo
 
-def ConstExpr.evalBranchTarget [Labels] (target : ConstExpr) (p : Std.Rco Int64) : Int64 :=
+-- Evaluate a PC-relative target expression (used by branches, `ADR`/`ADRP`).
+@[kstep, simp] def ConstExpr.evalTarget [Labels] (target : ConstExpr) (p : Std.Rco Int64) : Int64 :=
   match target with
   | .int64 imm => p.lower + imm
   | _ => target.interp p
@@ -413,14 +414,9 @@ structure StatusFlags.from_result.Remaining where
   val &&& BitVec.ofNat w.bits (w.bits - 1)
 
 @[kstep, simp] def maskOfLen {w : RegWidth} (len : Nat) : w.type :=
-  if len >= w.bits then
-    ~~~(0 : w.type)
-  else
-    ((1 : w.type) <<< len) - 1
+  ((1 : w.type) <<< len) - 1
 
 @[kstep, simp] def evalUBFM {w : RegWidth} (src : w.type) (immr imms : Nat) : w.type :=
-  let immr := immr % w.bits
-  let imms := imms % w.bits
   if imms >= immr then
     let len := imms - immr + 1
     let field := (src >>> immr).take len
@@ -432,8 +428,6 @@ structure StatusFlags.from_result.Remaining where
     (field.zeroExtend w.bits) <<< pos
 
 @[kstep, simp] def evalSBFM {w : RegWidth} (src : w.type) (immr imms : Nat) : w.type :=
-  let immr := immr % w.bits
-  let imms := imms % w.bits
   if imms >= immr then
     let len := imms - immr + 1
     let field := (src >>> immr).take len
@@ -445,8 +439,6 @@ structure StatusFlags.from_result.Remaining where
     (field.signExtend (w.bits - pos)).zeroExtend w.bits <<< pos
 
 @[kstep, simp] def evalBFM {w : RegWidth} (dst src : w.type) (immr imms : Nat) : w.type :=
-  let immr := immr % w.bits
-  let imms := imms % w.bits
   if imms >= immr then
     let len := imms - immr + 1
     let mask : w.type := maskOfLen (w := w) len
@@ -630,12 +622,12 @@ set_option maxHeartbeats 1000000
   | .SDIV dst src1 src2 =>
     let val1 := s.regs.getRegOrZr src1
     let val2 := s.regs.getRegOrZr src2
-    let res := if val2 == 0 then (0 : w.type) else val1.sdiv val2
+    let res := val1.sdiv val2
     s.setRegOrZr dst res next
   | .UDIV dst src1 src2 =>
     let val1 := s.regs.getRegOrZr src1
     let val2 := s.regs.getRegOrZr src2
-    let res := if val2 == 0 then (0 : w.type) else val1 / val2
+    let res := val1 / val2
     s.setRegOrZr dst res next
   | .SMADDL dst src1 src2 src3 =>
     let val1 := s.regs.getRegOrZr src1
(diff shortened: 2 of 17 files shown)


Kraken/AArch64/Syntax.lean
```diff
@@ -247,7 +247,7 @@ Shifted register operand for logical and arithmetic instructions (`Xm, <shift> #
 -/
 structure ShiftRegExpr (w : RegWidth) where
   reg : RegOrZr w
-  amount : Int64 -- Must be in the range $[0, 31]$ for 32-bit instructions and [0, 63] for 64-bit instructions.
+  amount : Nat -- Must be in the range $[0, 31]$ for 32-bit instructions and [0, 63] for 64-bit instructions.
```
miriampolzer, line 250: Why did you change this to an unbounded type? Is it easier to apply the shift operation on non bitvec or bitvec-like types?

Kraken/AArch64/Semantics.lean
```diff
@@ -199,15 +185,15 @@ def MachineData.load
   (s : MachineData) (addr : BitVec 64) (w : MemWidth)
   (ret : w.type → MachineData → Effects): Effects :=
   require_read_access addr w (fun _unit =>
-    match Mem.loadInt s.dmem addr w.bytes with
-    | .some i => ret (.ofInt _ i) s
+    match Mem.loadBV s.dmem addr w.bits w.bytes with
```
protz: can the bits argument be derived from the bytes argument?

Kraken/AArch64/Semantics.lean
```diff
@@ -221,29 +207,29 @@ export Labels (label)
...
-               | .UXTH => (v.take MemWidth.W16.bits).unsigned
-               | .SXTH => (v.take MemWidth.W16.bits).signed
-               | .UXTW => (v.take MemWidth.W32.bits).unsigned
-               | .SXTW => (v.take MemWidth.W32.bits).signed
-               | .UXTX => v.unsigned
-               | .SXTX => v.signed
+               | .UXTB => (v.take MemWidth.W8.bits).setWidth 64
+               | .SXTB => (v.take MemWidth.W8.bits).signExtend 64
+               | .UXTH => (v.take MemWidth.W16.bits).setWidth 64
+               | .SXTH => (v.take MemWidth.W16.bits).signExtend 64
```
protz, line 229: nice -- I like this very explicit style

Kraken/AArch64/Semantics.lean
```diff
@@ -402,25 +388,29 @@ structure StatusFlags.from_result.Remaining where
 
 @[kstep, simp] def StatusFlags.from_result {w} (result : BitVec w) (f : from_result.Remaining) : StatusFlags :=
   { n := result.msb
-    z := result == BitVec.zero _
+    z := result == 0#w -- bv_decide abstracts `BitVec.zero w` as an opaque variable
```
protz, line 392: is this a bug?

protz: as in: should `bv_decide` be aware of `BitVec.zero w`?

Kraken/SeparationMem.lean
```diff
@@ -96,19 +96,19 @@ theorem storeBytes_sep {w : Nat} (a : BitVec w) (n : Nat) (_bs bs : List UInt8)
   rw [sep_comm]
   exact ⟨m2, bs.At a, rfl, disjoint_symm (disjoint_Atsame_l_same_r _bs bs a m2 h_inter (by omega)), hR, rfl⟩
 
-theorem loadInt_sep {w : Nat} (bs : List UInt8) (a : BitVec w) (n : Nat) (R : Mem w → Prop) (m : Mem w)
+theorem loadBV_sep {w : Nat} (bs : List UInt8) (a : BitVec w) (bits n : Nat) (R : Mem w → Prop) (m : Mem w)
```
miriampolzer: Do I see it correctly that n is the number of bytes to load, and bits is the number of bits to load? Could we just pass one, e.g. keep n and calculate bits as `n*8`?

Kraken/BitVec.lean
```diff
@@ -0,0 +1,92 @@
...
+    c - a.setWidth 64 * b.setWidth 64
+      = BitVec.ofInt 64 (c.unsigned - a.unsigned * b.unsigned) := by
+  simp only [BitVec.unsigned]
+  rw [BitVec.ofInt_sub, ← BitVec.ofInt_mul_toNat (by omega),
+    BitVec.ofInt_toNat_setWidth, BitVec.setWidth_eq]
+
+-- x64 specific lemmas.
+
+-- `mul` and `mulx`
+theorem mul_widen_unsigned {w n : Nat} (hn : w + w ≤ n) (a b : BitVec w) :
```
protz, line 64: where are these theorems useful? should they be tagged with `simp`?

Kraken/Flags.lean
```diff
@@ -0,0 +1,242 @@
...
+
+    cf := v.unsigned != a.unsigned + b.unsigned
+    of := v.signed   != a.signed   + b.signed
+
+Here we define the same flags in `BitVec` terms. The idea is to widen by one bit
+and look at the carry position, or compare sign bits. The `BitVec` forms are what
+the semantics evaluate. The `_spec` theorems are provided for arithmetic reasoning
+that is more natural over `Int`.
+
+TODO: `bv_decide` does not reduce `BitVec.carry` and abstracts it as an
```
protz, line 17: should these be feature requests for Lean?

Kraken/AArch64/Parser.lean
```diff
@@ -951,8 +951,8 @@ def parseShiftRegExpr (w : RegWidth) (allowRor : Bool := false) : Parser (ShiftR
       | _ => fail s!"unknown shift type: {shiftName}"
     skipHWs
     let amt ← parseInt64
```
miriampolzer: Instead of parsing as Int64 and then checking and converting to Nat, could we just parse as Nat and only check the value?

Kraken/AArch64/Parser.lean
```diff
@@ -1856,16 +1856,16 @@ def parseCondCompare
     let imm ← parseImmNat
     liftExcept (checkCcmpImmediate imm)
     parseComma
-    let nzcv ← parseImmNat
-    liftExcept (checkNzcvFlags nzcv)
+    let nzcvNat ← parseImmNat
```
miriampolzer: Can you implement a bitvec parser instead that returns an error if the value does not fit into the given size? That would be consistent with existing parsing of e.g. Nat. and reduces the parse + check to just one parsing line.

Kraken/AArch64/Semantics.lean
```diff
@@ -402,25 +388,29 @@ structure StatusFlags.from_result.Remaining where
...
     c := f.c
     v := f.v }
 
 @[kstep, simp] def StatusFlags.adds {w} (res val1 val2 : BitVec w) : StatusFlags :=
   StatusFlags.from_result res {
-    c := res.unsigned != val1.unsigned + val2.unsigned
-    v := res.signed != val1.signed + val2.signed }
+    c := Kraken.Flags.addCarry val1 val2 false
```
miriampolzer, line 398: That's nice, now we have unified logic, and it's easier to read this because you don't need to scratch your head about the exact term, you just know that this is the carry flag from add.

Kraken/X64/Semantics.lean
```diff
@@ -280,14 +266,13 @@ def MachineData.load
...
+    match Mem.loadBV s.dmem addr w.bits w.bytes with
+    | .some v => ret v s
     | .none => nonmem_load s.dmem addr w (fun v dmem => ret v { s with dmem }))
 
--- Alternatively, we could define this in terms of BitVecs without %:
--- (addr &&& BitVec.ofNat 64 (bytes - 1)) == 0#64
+-- `bytes` is always a power of two, so alignment is a bit mask test.
 def isAligned (bytes : Nat) (addr : BitVec 64) : Bool :=
-  addr.toNat % bytes == 0
+  addr &&& BitVec.ofNat 64 (bytes - 1) == 0#64
```
miriampolzer, line 276: This modulo is a bit hard to read if you're not used to it, is there a simpler way to express this, e.g. through truncating and then zero extending the bitvector? (If this is standard practice, ignore my comment, I'll get used to it)

Kraken/X64/Semantics.lean
```diff
@@ -405,8 +391,24 @@ def CondCode.interp (cc : CondCode) (s : StatusFlags) : Bool := match cc with
...
+
+/-- The rotate amount for `rol`/`ror`, reduced modulo the operand width. -/
+@[kstep] def ShiftCountExpr.interpRotate [Labels] (c : ShiftCountExpr) (s : MachineData)
+    (p : Std.Rco Int64) (w : Width) : BitVec w.bits :=
+  (c.interpMaskedBV s p w).setWidth w.bits &&& BitVec.ofNat w.bits (w.bits - 1)
+
+/-- The rotate amount for `rcl`/`rcr`, reduced modulo `1 + w.bits`. -/
+@[kstep] def ShiftCountExpr.interpRotateCarry [Labels] (c : ShiftCountExpr) (s : MachineData)
+    (p : Std.Rco Int64) (w : Width) : BitVec (1 + w.bits) :=
+  (c.interpMaskedBV s p w).setWidth (1 + w.bits) % BitVec.ofNat (1 + w.bits) (1 + w.bits)
```
miriampolzer: Oh, so the bitvec type does have a modulo operation?

Kraken/X64/Semantics.lean
```diff
@@ -841,5 +838,5 @@ abbrev eval [layout : Layout] (prog : Program) := Executable.eval (layout prog)
     .instr (.regular .W64 .W64 (.inc (.reg (.low .rax .W64)))),
     .instr (.regular .W64 .W64 .ret) ]
   let start := (Executable.labels exe).label "main"
-  let data : MachineData := { dmem := Mem.storeInt {} 0x100 8 0x1337, regs := {rsp := 0x100} }
+  let data : MachineData := { dmem := Mem.storeBV {} 0x100 8 (0x1337 : BitVec 64), regs := {rsp := 0x100} }
```
miriampolzer: nit: In some places you did `<num>#w` in others you put the type. Is there a reason? E.g. would 0x1337#64 not work?

Kraken/Mem.lean
```diff
@@ -21,8 +21,9 @@ abbrev Mem (w) := ExtHashMap (BitVec w) UInt8
 def Mem.loadBytes {w} (m : Mem w) (a : BitVec w) (n : Nat) : Option (List UInt8) :=
   ((List.range n).map (fun i => m.get? (a + .ofNat _ i))).allSome
 
-def Mem.loadInt {w} (m : Mem w) (a : BitVec w) (n : Nat) : Option Int :=
-  (loadBytes m a n).map Int.ofBytes
+/-- Read `n` bytes at `a` and assemble them into a `bits`-wide value. -/
+def Mem.loadBV {w} (m : Mem w) (a : BitVec w) (bits n : Nat) : Option (BitVec bits) :=
```
miriampolzer: Same redundancy about bits/n here?

miriampolzer (commented):
I like what you did with the flags, now the logic is nicely unified! So far I have looked at changes you made to Semantics, Syntax, Examples and Parser. Still need to review the flags, bitvec and the stdlib candidates file.

Kraken/AArch64/Semantics.lean
lukaszobernig: Your question is related to https://github.com/AeneasVerif/kraken/pull/157#discussion_r3810753949. I deliberately kept `Mem.loadBV` generic without assuming that the number of bits is a multiple of 8. But maybe that is not so useful and we should just assume this. I have no strong opinion either direction here.

Kraken/AArch64/Semantics.lean
lukaszobernig: I would consider it a bug, yes.

Kraken/BitVec.lean
lukaszobernig: I am unsure if tagging them with simp at this point is useful. My reason for adding them is that bitwise reasoning about multiplications is not that great of an experience in general. But now everything is expressed in terms of `BitVec`. These lemmas have the correct shape to match the individual multiplies that occur in both the x64 and AArch64 semantics, so that a user can translate the `BitVec`-based forms back into integer arithmetic. And then hopefully will be able to use `lia` or something equivalent. But we need to experiment with that first to decide on whether it is actually useful.

Kraken/Flags.lean
lukaszobernig: Similar to https://github.com/AeneasVerif/kraken/pull/157#discussion_r3810733745, we should probably file these as bugs/feature requests.

Kraken/AArch64/Syntax.lean
lukaszobernig: Exactly, `BitVec.shiftLeft`, `BitVec.ushiftRight`, `BitVec.sshiftRight` all take their shift amounts as `Nat` and that saves use one `Int64` to `Nat` conversion.

Kraken/SeparationMem.lean
lukaszobernig: See https://github.com/AeneasVerif/kraken/pull/157#discussion_r3810728349. We could assume that `bits` is always a multiple of 8 at the expense of some generality. I have no strong opinion either direction here.

Kraken/Mem.lean
lukaszobernig: See https://github.com/AeneasVerif/kraken/pull/157#discussion_r3810728349 and https://github.com/AeneasVerif/kraken/pull/157#discussion_r3810753949.

Kraken/X64/Semantics.lean
lukaszobernig: Definitely, I'll change them all to `num#width` to make it consistent!

Kraken/X64/Semantics.lean
lukaszobernig: Yes, but on second thought this is not ideal. Let me replace this with something better.

Kraken/AArch64/Parser.lean
lukaszobernig: Good point! I'll change it as you suggested.

Kraken/AArch64/Parser.lean
lukaszobernig: Indeed, I should have just use `parseImmNat`, I'll fix this!

Kraken/X64/Semantics.lean
lukaszobernig: We could truncate to `log2 bytes` and then test if the truncation is zero. Since we have `w.bits = log2 bytes`, this should also be `bv_decide` friendly. On the other hand, testing whether something is aligned to a power-of-2 this way is pretty standard, see e.g. https://graphics.stanford.edu/~seander/bithacks.html#DetermineIfPowerOf2 or https://github.com/torvalds/linux/blob/master/tools/include/linux/align.h.

Kraken/AArch64/Semantics.lean
lukaszobernig: OK changed it so we always consider a number of bytes (and assume byte = 8 * bits).

Kraken/SeparationMem.lean
lukaszobernig: Removed `bits`, now we always consider a number of bytes.

Kraken/Mem.lean
lukaszobernig: Changed this too.

Kraken/X64/Semantics.lean
protz: I would just put the modulo as a comment above the optimized implementation for those of us who are not familiar with this

lukaszobernig: Sounds good, done!

Kraken/BitVec.lean
```diff
@@ -0,0 +1,92 @@
...
+    (a b : BitVec w) :
+    a.signExtend n * b.signExtend n = BitVec.ofInt n (a.signed * b.signed) :=
+  (BitVec.ofInt_mul_toInt hw hn a b).symm
+
+theorem setWidth_mul_setWidth_eq_ofInt {w n : Nat} (hn : w + w ≤ n) (a b : BitVec w) :
+    a.setWidth n * b.setWidth n = BitVec.ofInt n (a.unsigned * b.unsigned) :=
+  (BitVec.ofInt_mul_toNat hn a b).symm
+
+-- `SMADDL`/`UMADDL`, `SMSUBL`/`UMSUBL`
+theorem signExtend_mul_add_eq_ofInt (a b : BitVec 32) (c : BitVec 64) :
```
miriampolzer, line 41: These could also be solved with simp:
```
theorem signExtend_mul_add_eq_ofInt (a b : BitVec 32) (c : BitVec 64) :
    a.signExtend 64 * b.signExtend 64 + c
      = BitVec.ofInt 64 (a.signed * b.signed + c.signed) := by
  simp [signed, ofInt_add, ofInt_mul_toInt]

theorem setWidth_mul_add_eq_ofInt (a b : BitVec 32) (c : BitVec 64) :
    a.setWidth 64 * b.setWidth 64 + c
      = BitVec.ofInt 64 (a.unsigned * b.unsigned + c.unsigned) := by
  simp [unsigned, ofInt_add, ofInt_mul_toNat]

theorem sub_signExtend_mul_eq_ofInt (a b : BitVec 32) (c : BitVec 64) :
    c - a.signExtend 64 * b.signExtend 64
      = BitVec.ofInt 64 (c.signed - a.signed * b.signed) := by
  simp [signed, ofInt_sub, ofInt_mul_toInt]

theorem sub_setWidth_mul_eq_ofInt (a b : BitVec 32) (c : BitVec 64) :
    c - a.setWidth 64 * b.setWidth 64
      = BitVec.ofInt 64 (c.unsigned - a.unsigned * b.unsigned) := by
  simp [unsigned, ofInt_sub, ofInt_mul_toNat]
```

Kraken/StdLibCandidates.lean
```diff
@@ -0,0 +1,215 @@
...
+    ← Int.emod_emod_of_dvd _ hd, Int.bmod_emod, Int.emod_emod_of_dvd _ hd]
+  omega
+
+/-- Reducing an unsigned value modulo `2 ^ n` is truncation to `n` bits. Unlike
+`ofInt_toInt_setWidth` this needs no side condition. Both sides zero-extend when
+`n` exceeds the source width.
+
+The signed counterpart needs no lemma at all, `BitVec.signExtend` is defined
+as `ofInt v x.toInt`. -/
+theorem ofInt_toNat_setWidth {m n : Nat} (y : BitVec m) :
```
miriampolzer, line 32: `simp` can solve this

Kraken/StdLibCandidates.lean
```diff
@@ -0,0 +1,215 @@
...
+def rorBV {w : Nat} (x : BitVec w) (m : BitVec w) : BitVec w :=
+  (x >>> m) ||| (x <<< (BitVec.ofNat w w - m))
+
+/-- Rotate left by a bitvector amount, expressed with shifts. -/
+def rolBV {w : Nat} (x : BitVec w) (m : BitVec w) : BitVec w :=
+  (x <<< m) ||| (x >>> (BitVec.ofNat w w - m))
+
+theorem rolBV_eq_rotateLeft {w : Nat} (x m : BitVec w) (hm : m.toNat < w) :
+    x.rolBV m = x.rotateLeft m.toNat := by
+  have hwlt : w < 2 ^ w := Nat.lt_two_pow_self
```
miriampolzer: Is there an advantage to have clauses without any proofs/arguments? For me this can be reduced to:
```
theorem rolBV_eq_rotateLeft {w : Nat} (x m : BitVec w) (hm : m.toNat < w) :
    x.rolBV m = x.rotateLeft m.toNat := by
  have hwBV : (BitVec.ofNat w w).toNat = w := by
    rw [BitVec.toNat_ofNat, Nat.mod_eq_of_lt Nat.lt_two_pow_self]
  have hsub : ((BitVec.ofNat w w) - m).toNat = w - m.toNat := by
    rw [BitVec.toNat_sub, hwBV]
    have hm2 := m.isLt
    rw [show (2 ^ w - m.toNat) + w = 2 ^ w + (w - m.toNat) by omega,
      Nat.add_mod_left, Nat.mod_eq_of_lt (by omega)]
  simp [rolBV, BitVec.rotateLeft_def, Nat.mod_eq_of_lt hm, hsub]
```
Also, I've replaced a rw with simp, let me know if that makes sense. For me it looks a bit simpler, but maybe there is an advantage to using rw?

Kraken/StdLibCandidates.lean
```diff
@@ -0,0 +1,215 @@
...
+
+theorem rotateLeft_mod {w : Nat} (x : BitVec w) (n : Nat) :
+    x.rotateLeft (n % w) = x.rotateLeft n := by
+  simp [BitVec.rotateLeft_def]
+
+theorem rotateRight_mod {w : Nat} (x : BitVec w) (n : Nat) :
+    x.rotateRight (n % w) = x.rotateRight n := by
+  simp [BitVec.rotateRight_def]
+
+theorem rorBV_eq_rotateRight {w : Nat} (x m : BitVec w) (hm : m.toNat < w) :
```
miriampolzer, line 168: Same as the above, I think this can be simplified to:
```
theorem rorBV_eq_rotateRight {w : Nat} (x m : BitVec w) (hm : m.toNat < w) :
    x.rorBV m = x.rotateRight m.toNat := by
  have hwBV : (BitVec.ofNat w w).toNat = w := by
    rw [BitVec.toNat_ofNat, Nat.mod_eq_of_lt Nat.lt_two_pow_self]
  have hsub : ((BitVec.ofNat w w) - m).toNat = w - m.toNat := by
    rw [BitVec.toNat_sub, hwBV]
    have hm2 := m.isLt
    rw [show (2 ^ w - m.toNat) + w = 2 ^ w + (w - m.toNat) by omega,
      Nat.add_mod_left, Nat.mod_eq_of_lt (by omega)]
  simp [rorBV, BitVec.rotateRight_def, Nat.mod_eq_of_lt hm, hsub]
```

Kraken/X64/Examples/Examples.lean
```diff
@@ -398,21 +396,19 @@ theorem alu_mem_example_correct [layout : Layout] (s₀ : MachineData)
...
-  case h_len => exact Int.toBytes_length 8 _
+  case h_len => exact BitVec.toBytes_length 8 _
   kstep
   tactic =>
   apply Eventually.done
-  dsimp [UInt64.toBitVec]
-  change (100 : UInt64) + { toBitVec := BitVec.ofInt 64 (Int.ofBytes (Int.toBytes 8 (42#64).toInt)) } = (142 : UInt64)
-  rw [BitVec.ofInt_ofBytes_toBytes 64 8 rfl]
-  rfl
+  simp (config := { zetaDelta := true }) only [BitVec.ofBytes_toBytes 64 8 rfl]
```
miriampolzer: simp without only works for me too here, then you don't even need to call bv_decide.

miriampolzer (approved):
I looked at the remaining files now, tried them out myself, and proposed a few simplificiations. The rest looks good to me!
I really like what you did with the flags, the specs make it all nice and clean and gives confidence that we're computing the righ thing.

Kraken/BitVec.lean
lukaszobernig: Thanks, looks much nicer now!

Kraken/StdLibCandidates.lean
lukaszobernig: Done!

Kraken/StdLibCandidates.lean
lukaszobernig: No using simp is definitely fine!

Kraken/StdLibCandidates.lean
lukaszobernig: Done!

Kraken/X64/Examples/Examples.lean
lukaszobernig: Nice, we were also missing an annotation for `AddressSize.address_size` and now a single `rfl` closes it.

Kraken/AArch64/Semantics.lean
protz: did you end up submitting this upstream?

protz (approved):
Looks good to me, thank you, this is a nice simplification to have

andres-erbsen:
There seem to be two changes here:
- replacing UInt64 with BitVec 64 (and similar)
- replacing Stdlib arithmetic for flag computations with custom definitions in terms of bitvectors

I can see how each would improve the applicability of bv_decide, but I worry that the latter would be a significant barrier to applicability of lia and thus grind. Perhaps support for the new definitions could be added through grind attributes or similar, but replacing the (imo straightforward) arithmetic formulations of concepts such as overflow with bit-blasting formulas in the specification does not seem satisfying to me. If we need to factor things to make bv_decide work, I think we should still keep the old definitions, just behind a name that bv_decide can have a lemma for. I am in general more long-term-optimistic about grind rather than bv_decide as a central automation tool, so deeply changing the *Semantics* to make bitblasting easier seems like a step in the wrong direction.

I am relatively ambivalent about the UInt->BitVec change, though I wonder whether it affects performance of interpreter usage and thus testing.

andres-erbsen:
FYI, I a vibe-ported version of my intblasting test suite from https://github.com/AeneasVerif/kraken/pull/122 on latest Lean is here: https://github.com/leanprover/lean4/pull/15224

andres-erbsen:
Update: I have opened https://github.com/leanprover/lean4/issues/15260 and https://github.com/leanprover/lean4/pull/15259

Edit: and https://github.com/leanprover/lean4/issues/15261 and https://github.com/leanprover/lean4/pull/15262

Issue #148, opened 2026-08-10
Use the module system to fail faster in separation logic unification goals


protz:
This was my repro. The goal is not to make this succeed, the goal is to make it fail fast -- right now, even a small typo or a mismatch causes the unified to not return, which makes for a subpar interactive proof experience.

```lean4
import Std.Data.ExtHashMap

open Std
open Std.ExtHashMap

def sep (p q : ExtHashMap (BitVec 64) UInt8 → Prop) (m : ExtHashMap (BitVec 64) UInt8) : Prop :=
  ∃ a b, a.union b = m ∧ a.inter b = ∅ ∧ p a ∧ q b

infixl:60 " ⋆ " => sep

def atBytes (bs : List UInt8) (a : BitVec 64) : ExtHashMap (BitVec 64) UInt8 :=
  .ofList (bs.mapIdx fun i b => (a + BitVec.ofNat 64 i, b))

example (bs : List UInt8) (addr : BitVec 64) (R : ExtHashMap (BitVec 64) UInt8 → Prop) (m : ExtHashMap (BitVec 64) UInt8)
    (h : (Eq (atBytes [0, 0, 0, 0, 0, 0, 0, 0] addr) ⋆ (Eq (atBytes bs (addr + 8#64)) ⋆ R)) m) :
    ∃ R', (Eq (atBytes bs (addr + 8#64)) ⋆ R') m := by
  exact ⟨_, h⟩
```

Sebastian's suggestion was:

> I think we're getting close to the point where we do just want to recommend the use of the module system in general, and indeed you can now have Lean recommend that to all your downstream users by setting [requiresModuleSystem](https://lean-lang.org/doc/reference/latest/Build-Tools-and-Distribution/Lake/#Lake___PackageConfig-requiresModuleSystem). Is that a potential solution for you, and is it sufficient?

So, switching Kraken to use the module system might bring us better performance.

protz:
Real-world example: removing the `simp` here: https://github.com/AeneasVerif/kraken/blob/93f33808eccf45e771e7c874a5452a5263ffb99d/Kraken/X64/Examples/Examples.lean#L359 causes a fair amount of grief in unification

protz:
Discussion with @joehattori: let's put this under regression as part of #158, by setting a low number of heartbeats and making sure that this fails fast using #guard_msgs

PR #158, opened 2026-09-14
Adopt Lean's module system

joehattori:
- Require the module system throughout the codebase.
- Move the parser implementations into Parser/Basic.lean so the same ordinary definitions can be imported at both runtime and meta phases.
- Remove BitVec.reduceOfInt, Int.reduceBmod, Int.reduceNeg, Int64.reduceToInt, and Nat.reducePow from the ksimp registrations in both example suites. Their hidden implementation bodies make those registrations invalid under the module system. kstep continues to evaluate concrete arithmetic through the existing Sym.Simp.evalGround step, so the capability of ksimp remains the same.


protz (commented):
Thanks! I am starting with some high-level questions.
1. Why the new `Basic.lean` files in `Parser`?
2. It seems like all of your module imports are `public` -- from the documentation:

> The imported module's public scope is added to the current module's public scope and made available to the current module's importers.

This confuses me. For instance, in Parser/Basic.lean:

```
public import Std.Internal.Parsec.String
```

it seems like whoever does `import AArch64.Parser.Basic` will also bring into their scope the parser combinators. Is this what we want? 

3. I am also confused about the `@[expose] public section` in e.g. Basic.lean -- do we really need to expose the definitions of parser combinators to proofs? Are we going to use these in `simp` and/or `grind`?

My questions may be due to misunderstandings of how the Lean module system works, but it would be good to clarify in this PR so that we understand the intent of this redesign. Thanks!

joehattori:
Thanks for the review.

> Thanks! I am starting with some high-level questions.
> 
> 1. Why the new `Basic.lean` files in `Parser`?

It is to allow parser to be imported at both the runtime and the meta phase. I added a comment to the `Basic.lean` files.

> 2. It seems like all of your module imports are `public` -- from the documentation:
> 
> > The imported module's public scope is added to the current module's public scope and made available to the current module's importers.
> 
> This confuses me. For instance, in Parser/Basic.lean:
> 
> ```
> public import Std.Internal.Parsec.String
> ```
> 
> it seems like whoever does `import AArch64.Parser.Basic` will also bring into their scope the parser combinators. Is this what we want?

Indeed, some public imports were unnecessary. I removed some of the public imports in the latest commit. Please note that the core layers (e.g. Syntax, Semantics, Sep) still use public import because any module whose types or definitions appear in a public declaration's signature must be `public import`ed. 

> 
> 3. I am also confused about the `@[expose] public section` in e.g. Basic.lean -- do we really need to expose the definitions of parser combinators to proofs? Are we going to use these in `simp` and/or `grind`?

Thanks for pointing this out. I removed the lines from the `Basic.lean` files. 

> 
> My questions may be due to misunderstandings of how the Lean module system works, but it would be good to clarify in this PR so that we understand the intent of this redesign. Thanks!

protz (approved):
Thanks Joe! This is in good shape and I believe fixes #148 with your nice regression test.

